Issue #10, opened 2019-09-06
Provide `dotnet new` template for Web API


tomkerkhove:
Provide a [custom `dotnet new`](https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates) template which configures a Web API with minimal requirements.

The template should be calles `arcus-webapi`.

## Features in the template

- [X] Plug in [exception middleware](https://webapi.arcus-azure.net/features/logging)
- [X] Configure content negotiation to only support `application/json`
- [x] Setup Swagger generation
- [x] Setup Swagger UI (but disable for PROD env)
- [X] Configure [health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2) *(middleware only)*
- [X] Provide health endpoint *(based on [info gained via health checks](https://www.codit.eu/blog/documenting-asp-net-core-health-checks-with-openapi/)*
	- Fully documented according to our guidelines
- [x] Provide a working Dockerfile
- [X] Build a default logger

The template should not provide any authentication by default but should provide a comment that warns about this and point to [Arcus WebAPI and Shared Access Key](https://webapi.arcus-azure.net/features/security/auth/shared-access-key).

## Requirements

We should provide a CI that performs following steps:
1. Create project based on template via `dotnet new arcus-webapi`
2. Build a Docker container based on `Dockerfile`
3. Run Docker container
4. Get the health of the container

## Good reads

- https://rehansaeed.com/custom-project-templates-using-dotnet-new/

tomkerkhove:
I'm assigning this to you if that is ok @stijnmoreels 	.

However, let's take it step by step and provide the boilerplate with just the exception handling, for example, and already create a PR.

It's added to our Arcus roadmap but let's finalize the items that we've discussed today 
 first if that's ok.

tomkerkhove:
@stijnmoreels If I understand it correctly we are good to go, except for the correlation.

Is that correct?

If so, would be good if we started working on the NuGet release to ship a v0.1.

stijnmoreels:
Yes, besides correlation all the tasks are done. A closer look at the title, description, summary, icon, ... would be a good idea though.

What else must be done for the NuGet release?

tomkerkhove:
> A closer look at the title, description, summary, icon, ... would be a good idea though.

What are you referring to here?

tomkerkhove:
> What else must be done for the NuGet release?

[...]

PR #15, opened 2019-09-11
FEAT: add initial template proj with exception handling

stijnmoreels:
Initial Web API template project with exception handling.

The project also adds two warnings about authentication and HTTPS configuration not being configured yet.

Relates to #10


stijnmoreels:
Not sure what to do with the `appsettings.Development.json` file. Should I delete this?

stijnmoreels:
The requirements talk about asserting on the health of the project; but since we discussed to start with the exception handling; I'm not sure if I already can fully test this?

tomkerkhove:
Since they will have a Dockerfile I would suggest to remove the `appsettings.json` since they will not be used - What do you think?

PR #16, opened 2019-09-18
FEAT: add health check via middleware

stijnmoreels:
Relates to #10

### DIFF
--- a/src/Arcus.Template.WebApi/Startup.cs
+++ b/src/Arcus.Template.WebApi/Startup.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
 
 namespace Arcus.Template.WebApi
 {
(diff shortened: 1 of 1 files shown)


stijnmoreels:
Not sure though if this middleware should be removed when we have a full health endpoint?

src/Arcus.Template.WebApi/Startup.cs
```diff
@@ -22,6 +26,8 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             #warning "Please configure application with authentication mechanism: https://webapi.arcus-azure.net/features/security/auth/shared-access-key"
 
             app.UseMvc();
+
+            app.UseHealthChecks("/health");
```
tomkerkhove: We don't really need this as we'll expose the endpoint ourselves based on the health middleware. See https://www.codit.eu/blog/documenting-asp-net-core-health-checks-with-openapi/ for more info

src/Arcus.Template.WebApi/Startup.cs
```diff
@@ -11,6 +12,9 @@ public class Startup
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddMvc();
+
+            services.AddHealthChecks()
+                    .AddCheck("self", () => HealthCheckResult.Healthy());
```
tomkerkhove: What happens if we leave this one out?

tomkerkhove:
@stijnmoreels 	We will keep using the middleware, but not expose it via the built-in capabilities. See  codit.eu/blog/documenting-asp-net-core-health-checks-with-openapi  why

src/Arcus.Template.WebApi/Startup.cs
fgheysels: I don't think you need the 'self' check.

PR #17, opened 2019-09-18
FEAT: only allow 'JsonInputFormatter'

stijnmoreels:
Relates to #10


stijnmoreels:
Not sure to also restrict the response to only json?

PR #18, opened 2019-09-18
FEAT: add console logging as default logger

stijnmoreels:
Relates to #10


stijnmoreels:
We could register a logger that discards the log messages. That way we don't take dependency on the log console NuGet package?

tomkerkhove:
@stijnmoreels 	What do you mean? Can you give an example please?

_PS: I have no issue with the dependency that was added to be honest._

stijnmoreels:
> @stijnmoreels What do you mean? Can you give an example please?
> 
> _PS: I have no issue with the dependency that was added to be honest._

Just a thought: if we setup the logging system for the API but register a provider that discards the log messages (ex. `NullLogger`) than we don't need any extra dependency and yet have shown the user to provide a logging system.

tomkerkhove:
That is correct but the idea is that the template provides a easy to setup and good to get started template with the minimal requirements so we should have console logging at least!

PR #20, opened 2019-09-18
FEAT: add health endpoint to default API project

stijnmoreels:
Relates to #10

### DIFF
--- a/src/Arcus.Template.WebApi/Controllers/HealthController.cs
+++ b/src/Arcus.Template.WebApi/Controllers/HealthController.cs
@@ -29,7 +29,7 @@ namespace Arcus.Template.WebApi.Controllers
         /// <summary>
         ///     Get health of API application.
         /// </summary>
-        /// <remarks>Provides an indication about the health of the API and its dependencies (Sim Database).</remarks>
+        /// <remarks>Provides an indication about the health of the API.</remarks>
         /// <response code="200">API is healthy</response>
         /// <response code="503">API is unhealthy or in degraded state</response>
         [HttpGet]
--- a/src/Arcus.Template.WebApi/Controllers/HealthController.cs
+++ b/src/Arcus.Template.WebApi/Controllers/HealthController.cs
@@ -3,7 +3,6 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Diagnostics.HealthChecks;
 using GuardNet;
-using Swashbuckle.AspNetCore.Annotations;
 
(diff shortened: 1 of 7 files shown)


src/Arcus.Template.WebApi/Controllers/HealthController.cs
```diff
@@ -0,0 +1,52 @@
...
+using GuardNet;
+
+namespace Arcus.Template.WebApi.Controllers
+{
+    /// <summary>
+    /// API endpoint to check the health of the application.
+    /// </summary>
+    [ApiController]
+    [Route("api/v1/[controller]")]
+    public class HealthController : ControllerBase
```
tomkerkhove: Let's adhere to https://github.com/CoditEU/practical-api-guidelines/tree/master/maturity-level-one#document-your-apis

For example:

- Summary should just be `Get Health`
- OperationId `Health_Get` should be used

src/Arcus.Template.WebApi/Controllers/HealthController.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using GuardNet;
+
+namespace Arcus.Template.WebApi.Controllers
+{
+    /// <summary>
+    /// API endpoint to check the health of the application.
+    /// </summary>
+    [ApiController]
+    [Route("api/v1/[controller]")]
```
tomkerkhove: I'm not a fan of this, but can just be me :D

The reason is that if we rename the controller it will automatically break our API contract. Because of this I always explicitly add the route
```suggestion
    [Route("api/v1/health")]
```

src/Arcus.Template.WebApi/Controllers/HealthController.cs
```diff
@@ -27,14 +28,15 @@ public HealthController(HealthCheckService healthCheckService)
...
-        ///     Get health of API application.
+        ///     Get Health
         /// </summary>
         /// <remarks>Provides an indication about the health of the API.</remarks>
         /// <response code="200">API is healthy</response>
         /// <response code="503">API is unhealthy or in degraded state</response>
         [HttpGet]
         [ProducesResponseType(typeof(HealthReport), StatusCodes.Status200OK)]
         [ProducesResponseType(typeof(HealthReport), StatusCodes.Status503ServiceUnavailable)]
+        [SwaggerOperation(OperationId = "Health_Get")]
```
tomkerkhove: Let's move it to `[HttpGet(Name= "Health_Get")]` instead similar to in the link, if that's ok?

PR #21, opened 2019-09-20
FEAT: CI YAML build w/ integration test running on Docker container

stijnmoreels:
- Added a new integration test that checks the previously created health endpoint
- Added Docker file that creates a new .NET project from the template project and uses that as entrypoint
- A new YAML CI build file is created to run the integration test after running the Docker container

Relates to #10


src/Arcus.Template.WebApi/Program.cs
```diff
@@ -15,7 +15,10 @@ public static void Main(string[] args)
 
         public static IWebHostBuilder CreateWebHostBuilder(string[] args)
         {
+            const string httpEndpointUrl = "http://+:5000";
```
stijnmoreels: Just a suggestion, not sure but do I have to retrieve this from the configuration?

stijnmoreels:
Example of test run: https://dev.azure.com/codit/Arcus/_build/results?buildId=37998&view=results

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2-stretch AS build
+WORKDIR /src
+COPY ["Arcus.Template.WebApi/Arcus.Template.WebApi.csproj", "Arcus.Template.WebApi/"]
+RUN dotnet restore "Arcus.Template.WebApi/Arcus.Template.WebApi.csproj"
+COPY . .
+WORKDIR "/src/Arcus.Template.WebApi"
+RUN dotnet build "Arcus.Template.WebApi.csproj" -c Release
+RUN dotnet new -i "bin/Release/Arcus.Template.WebApi.*.nupkg"
+RUN dotnet new arcus-webapi -n MyProject -o /output
```
stijnmoreels: 'MyProject' => just a name to create a new project.

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2-stretch AS build
+WORKDIR /src
+COPY ["Arcus.Template.WebApi/Arcus.Template.WebApi.csproj", "Arcus.Template.WebApi/"]
+RUN dotnet restore "Arcus.Template.WebApi/Arcus.Template.WebApi.csproj"
+COPY . .
+WORKDIR "/src/Arcus.Template.WebApi"
+RUN dotnet build "Arcus.Template.WebApi.csproj" -c Release
+RUN dotnet new -i "bin/Release/Arcus.Template.WebApi.*.nupkg"
```
stijnmoreels: A NuGet package is created for each template project so we can ship/test independenly.

build/ci-build.yml
```diff
@@ -0,0 +1,107 @@
...
+  paths:
+    include:
+      - src/*
+      - build/ci-build.yml
+
+variables:
+  - group: 'Build Configuration'
+  # Always use fixed version for .NET Core SDK
+  - name: 'DotNet.Sdk.Version'
+    value: '2.2.105'
```
tomkerkhove: This is .NET Core 2.2.3, we might want to use 2.2.7 instead.

build/ci-build.yml
```diff
@@ -0,0 +1,107 @@
...
+          vmImage: 'ubuntu-16.04'
+        steps:
+          - task: DotNetCoreInstaller@0
+            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.Version))'
+            inputs:
+              version: '$(DotNet.Sdk.Version)'
+          - task: DotNetCoreCLI@2
+            displayName: 'Compile'
+            inputs:
+              projects: 'src/*.sln'
```
tomkerkhove: Is it correct to assume we are only building the codebase here to run the tests given Dockerfile should build the required sources?

If so, I'd suggest to have only one stage to build the container, spin it up and run dotnet test on the codebase which will first run it anyway.

Would that make sense?

src/Arcus.Template.WebApi/Program.cs
tomkerkhove: If we use this, then it has to be passed via an environment variable.

This is more of a deployment thingy than fixed imo, no?

tomkerkhove: In terms of looking it up you can directly call environment variable or create default config provider and pass it to the builder as well.

Adding default configuration with environment variables would be good I think, no?

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.2-stretch-slim AS base
```
tomkerkhove: We should not be using strecht but alpine instead as it's smaller and more secure
```suggestion
FROM mcr.microsoft.com/dotnet/core/aspnet:2.2.7-alpine3.9 AS base
```

Let's pin to a specific version while we're at it.

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2-stretch AS build
```
tomkerkhove: Same here
```suggestion
FROM mcr.microsoft.com/dotnet/core/sdk:2.2.402-alpine3.9 AS build
```

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: 👌

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /src
+COPY ["Arcus.Template.WebApi/Arcus.Template.WebApi.csproj", "Arcus.Template.WebApi/"]
+RUN dotnet restore "Arcus.Template.WebApi/Arcus.Template.WebApi.csproj"
+COPY . .
```
tomkerkhove: We don't really need to copy everything I think?

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: It feels like the build part is two-fold:
- Build our template to generate a new NuGet package
- Build a new instance of our template, based on the NuGet package

Once that is done, we publish the new project.

Wouldn't it be better to have to stages:
- "Template Package Creation"
- "Project Generation & Publish", based on "Template Package Creation"

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR "/src/Arcus.Template.WebApi"
+RUN dotnet build "Arcus.Template.WebApi.csproj" -c Release
+RUN dotnet new -i "bin/Release/Arcus.Template.WebApi.*.nupkg"
+RUN dotnet new arcus-webapi -n MyProject -o /output
+WORKDIR "/output"
+RUN dotnet build -c Release -o /app
```
tomkerkhove: We can actually skip building as publish builds as well
```suggestion
RUN dotnet publish "MyProject.csproj" -c Release -o /app
```

src/docker-compose.override.yml
```diff
@@ -0,0 +1,9 @@
+version: '3.4'
+
+services:
+  arcus.template.webapi:
+    environment:
+      - ASPNETCORE_ENVIRONMENT=Development
+      - ARCUS_HTTP_PORT=5000
```
tomkerkhove: Huh, I probably missed that but where is this being used? You still need to add that, right?

src/Arcus.Template.Tests.Integration/Endpoints/v1/HealthEndpointTests.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        [Fact]
+        public async Task Health_Get_Succeeds()
+        {
+            // Act
+            using (HttpResponseMessage response = await _healthService.GetAsync())
+            {
+                // Assert
+                Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
tomkerkhove: We might want to check if the response contract according to what we expect?

src/Arcus.Template.Tests.Integration/Endpoints/HealthEndpointService.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        /// <summary>
+        /// Sends a GET request to the health endpoint of the API web application.
+        /// </summary>
+        public async Task<HttpResponseMessage> GetAsync()
+        {
+            _outputWriter.WriteLine("GET {0} ->", _healthEndpoint);
+            var response = await HttpClient.GetAsync(_healthEndpoint);
+            _outputWriter.WriteLine("<- {0}", response.StatusCode);
```
tomkerkhove: Please remind me to include correlation info when it's available

tomkerkhove:
Happy to discuss in person if you want @stijnmoreels !

build/ci-build.yml
stijnmoreels: Ah, yes. Now you remind, it was indeed something I saw in the logs.

Will make the YAML bit simpler. Thanks!

src/Arcus.Template.Tests.Integration/Endpoints/v1/HealthEndpointTests.cs
stijnmoreels: Yes, have to check how this is shown.

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: By stages, you mean YAML stages, therefore two docker files? Not entirely sure what you mean?

build/ci-build.yml
tomkerkhove: My pleasure!

src/Arcus.Template.Tests.Integration/Endpoints/v1/HealthEndpointTests.cs
tomkerkhove: What do you mean exactly? Just doing `.Content.ReadAs<HealthCheckResult>` (or what is it) and that it's not null is ok for me!

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: No a Docker stage - Everytime a Dockerfile has a `FROM` it is the start of a new stage - https://docs.docker.com/develop/develop-images/multistage-build/

But we can talk about this tomorrow if you want.

src/Arcus.Template.Tests.Integration/Endpoints/HealthEndpointService.cs
stijnmoreels: Ok, that would indeed be better.

build/ci-build.yml
stijnmoreels: :+1:

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: Will publish not only build the app when there is source code changes?

src/Arcus.Template.Tests.Integration/Endpoints/v1/HealthEndpointTests.cs
stijnmoreels: Done

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: Either retore+publish or build+publish

tomkerkhove:
As agreed we will keep the Dockerfile simple so that it just allows you to run the Web API.

To verify the whole chain we will change the CI so that we have the following stages:
- **Build**
  - Publish our template as a NuGet package
  - Publish package as artifact
- **Verification / Test**
  - Pull artifact from build
  - Install template
  - Run `dotnet new` for our template
  - Docker Build
  - Docker Run
  - Run integration tests
- **Publish**
  - Push the NuGet package to our MyGet feed (this could also be done in Build if you want)

src/Arcus.Template.WebApi/Program.cs
stijnmoreels: Done

build/ci-build.yml
stijnmoreels: Outdated: Dockerfile doesn't build this anymore

stijnmoreels:
>   * Push the NuGet package to our MyGet feed (this could also be done in Build if you want)

No other CI build in the Arcus repo pushes this to MyGet. Only Release YAML builds. Not sure why this should be added to the CI build and not the release build?

src/Arcus.Template.WebApi/Arcus.Template.WebApi.csproj
```diff
@@ -6,6 +6,10 @@
     <Company>Arcus</Company>
     <RepositoryType>Git</RepositoryType>
     <Description>Provide a template to easily build Web APIs running in Azure.</Description>
+    <IsPackable>true</IsPackable>
+    <PackageId>Arcus.Template.WebApi</PackageId>
```
tomkerkhove: `Template` or `Templates`?
```suggestion
    <PackageId>Arcus.Templates.WebApi</PackageId>
```

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.2.7-alpine3.9 AS base
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2-stretch AS build
+WORKDIR /src
+COPY ["Arcus.Template.WebApi.csproj", ""]
+RUN dotnet restore "./Arcus.Template.WebApi.csproj"
```
tomkerkhove: We can skip this and just use `dotnet publish` who implicitly does this

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY . .
+WORKDIR "/src/."
+RUN dotnet build "Arcus.Template.WebApi.csproj" -c Release -o /app
```
tomkerkhove: We can skip this and just use `dotnet publish` who implicitly does this

build/ci-build.yml
```diff
@@ -0,0 +1,117 @@
...
+  paths:
+    include:
+      - src/*
+      - build/ci-build.yml
+
+variables:
+  - group: 'Build Configuration'
+  # Always use fixed version for .NET Core SDK
+  - name: 'DotNet.Sdk.Version'
+    value: '2.2.206'
```
tomkerkhove: We are using 2.2.7 (right?), so this should be `2.2.402`

PS: We should move this a variable group given we need this in CI build + release build. If we do that, I'd create a group per Arcus project.

build/ci-build.yml
```diff
@@ -0,0 +1,117 @@
...
+          - task: DotNetCoreInstaller@0
+            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.Version)) in web API template'
+            inputs:
+              version: '$(DotNet.Sdk.Version)'
+          - task: DotNetCoreCLI@2
+            displayName: 'Install template'
+            inputs:
+              command: 'custom'
+              custom: 'new'
+              arguments: '-i "src/$(Project).WebApi/bin/Release/*.nupkg"'
```
tomkerkhove: Wouldn't it be better if our artifact is more structured and make the depending stages dumber?

Just thinking out loud, but the build artifact could be as following:
- `/templates/{namespace}.WebApi.nupkg` (later on extended with worker)
- `/tests/web-api`

build/ci-build.yml
```diff
@@ -0,0 +1,117 @@
...
+              imageName: 'myproject:$(Build.BuildId)'
+          - task: Docker@1
+            displayName: 'Run new web API project Docker image from template'
+            inputs:
+              command: 'Run an image'
+              imageName: 'myproject:$(Build.BuildId)'
+              containerName: 'myproject'
+              ports: '5000:5000'
+              envVars: |
+                ARCUS_HTTP_PORT=5000
```
tomkerkhove: We are explicitly specifying `5000` 4 times, how about a variable?

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Publish does everything that it should to output the artifact so that should be good to go

tomkerkhove:
> > * Push the NuGet package to our MyGet feed (this could also be done in Build if you want)
> 
> No other CI build in the Arcus repo pushes this to MyGet. Only Release YAML builds. Not sure why this should be added to the CI build and not the release build?

Now that we are part of the preview you can ignore my comment and just push it to GitHub Package Registry.

This should allow you to get started - https://github.com/tomkerkhove/github-package-registry-sandbox#nuget

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Resolving as it's no longer applicable

src/Arcus.Template.WebApi/Arcus.Template.WebApi.csproj
stijnmoreels: Ah, the project calls `Arcus.Template`. Since I find it more logical if you select a project:
`Arcus.Template.WebApi`.

build/ci-build.yml
stijnmoreels: :+1:

build/ci-build.yml
stijnmoreels: Cool idea. But then we have to split the build artifacts with the nuget packages otherwise the integration test project gets compiled twice.

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: I'm not sure that's entirely correct since the build fails with missing assembly files.

build/ci-build.yml
stijnmoreels: Yep, updated.

stijnmoreels:
> > > * Push the NuGet package to our MyGet feed (this could also be done in Build if you want)
> > 
> > 
> > No other CI build in the Arcus repo pushes this to MyGet. Only Release YAML builds. Not sure why this should be added to the CI build and not the release build?
> 
> Now that we are part of the preview you can ignore my comment and just push it to GitHub Package Registry.
> 
> This should allow you to get started - https://github.com/tomkerkhove/github-package-registry-sandbox#nuget

Same question: shoud this be in the CI build or rather in the release build? If in the CI build; we should then also add a 'Determine Package Version' task to this?

src/Arcus.Template.WebApi/Arcus.Template.WebApi.csproj
tomkerkhove: Sounds good to me!

build/ci-build.yml
tomkerkhove: Frankly, if you would just do `dotnet test {name}.Tests.integration.csproj` that will build it as well so that's fine for me as well

build/ci-build.yml
```diff
@@ -0,0 +1,119 @@
...
+  - group: 'Build Configuration'
+  # Always use fixed version for .NET Core SDK
+  - name: 'DotNet.Sdk.Version'
+    value: '2.2.402'
+  - name: 'Project'
+    value: 'Arcus.Template'
+  - name: 'Http.Port'
+    value: '5000'
+  - name: 'Arcus_Api_BaseUrl'
+    value: 'http://localhost:5000/api/v1'
```
tomkerkhove: You should use the variable here as well

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Based on https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish I don't see the need to restore (explicitly mentioned) and given it has a `--no-build` option it's clear that it builds it anyway. What is your concern?

stijnmoreels: My concert is that it doesn't build :).
I thought that publish was not exactly the same as building since publish will only build if any changes to the source was made and build will always re-compile the source.

stijnmoreels: Try it otherwise locally on any project; I get the same result.

build/ci-build.yml
stijnmoreels: Done.

build/ci-build.yml
stijnmoreels: Ok, done.

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: I'm not sure what your concern is, the docs clearly indicate that you can flag it to not build and that the output is ready to run:

> The dotnet publish command's output is ready for deployment to a hosting system (for example, a server, PC, Mac, laptop) for execution.

It doesn't mention that a build up front is required?

> Try it otherwise locally on any project; I get the same result.

What do you mean with same result?

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Can you incorporate this one as well please?

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: If I try to publish a non-build-yet project locally it also fails.
I know that the docs says otherwise, but apparently I can't get the expected result (even on a fresh project that's not build for the first time).

Some question/discussions online also talks about this. There is a diff between the two apparently.

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: It's ok

stijnmoreels:
> > > > * Push the NuGet package to our MyGet feed (this could also be done in Build if you want)
> > > 
> > > 
> > > No other CI build in the Arcus repo pushes this to MyGet. Only Release YAML builds. Not sure why this should be added to the CI build and not the release build?
> > 
> > 
> > Now that we are part of the preview you can ignore my comment and just push it to GitHub Package Registry.
> > This should allow you to get started - https://github.com/tomkerkhove/github-package-registry-sandbox#nuget
> 
> Same question: shoud this be in the CI build or rather in the release build? If in the CI build; we should then also add a 'Determine Package Version' task to this?

OK, that's done I think. I even used the template from the other repo.

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Ok, fine for me than - Sorry about that!

Can we then just remove the restore if it's ok - Otherwise just proceed and resolve the comments

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: Yeah I thought so too, but apperently with only `publish` in the Dockerfile, the build fails: non `.dll` file found.

The restore is run though (with `publish`) but no building of the source happen.

I checked this locally on a fresh project; and there I also got the same result: `publish` does `restore` but no `build`.

src/Arcus.Template.WebApi/Dockerfile
stijnmoreels: Ok, I'll remove the restore.

stijnmoreels: And just checked; the current Dockerfile (without restore) succeeds.

tomkerkhove: Awesome, thanks!

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Resolving as you are correct

src/Arcus.Template.WebApi/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.2.7-alpine3.9 AS base
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2.402-alpine3.9 AS build
+WORKDIR /src
+COPY ["Arcus.Template.WebApi.csproj", ""]
+
```
tomkerkhove: Nitpicking - Can you remove this new line please?

tomkerkhove:
I found this [build 🏗](https://dev.azure.com/codit/Arcus/_build?definitionId=765&_a=summary), which looks great, can you tweak the name and configure it to use the Arcus repo after merging please?

src/Arcus.Template.WebApi/Dockerfile
tomkerkhove: Nevermind, not going to block this PR on this.

tomkerkhove:
Nice job, love the CI!

stijnmoreels:
> I found this [build 🏗](https://dev.azure.com/codit/Arcus/_build?definitionId=765&_a=summary), which looks great, can you tweak the name and configure it to use the Arcus repo after merging please?

Yep, is done!

PR #22, opened 2019-10-02
FEAT: add swagger gen & UI only for development

stijnmoreels:
Adding swagger UI (as minimal as possible) to the Arcus web API template.

#10


src/Arcus.Template.WebApi/.template.config/template.json
```diff
@@ -22,6 +22,22 @@
...
+      "replaces": "//#if DEBUG"
+    },
+    "EndIf": {
+      "type": "generated",
+      "generator": "constant",
+      "parameters": {
+        "value": "#endif"
+      },
+      "replaces": "//#endif"
+    },
```
stijnmoreels: Because we don't use env app settings; we can't use `env.IsDevelopment()`; which means directives (`#if`) has to be used.
The problem was that this is used in the template creation and will be removed at creation time.
The fix is to comment these out and uncomment them on creation (in `template.json`).

src/Arcus.Template.WebApi/Startup.cs
```diff
@@ -61,6 +78,15 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             #warning Please configure application with authentication mechanism: https://webapi.arcus-azure.net/features/security/auth/shared-access-key
 #endif
             app.UseMvc();
+
+//#if DEBUG
+            app.UseSwagger();
+            app.UseSwaggerUI(swaggerUiOptions =>
+            {
+                swaggerUiOptions.SwaggerEndpoint("v1/swagger.json", "Arcus.Template.WebApi");
+                swaggerUiOptions.DocumentTitle = "Arcus.Template.WebApi";
```
tomkerkhove: Can we plug in the name of the project that was created based on this template?

src/Arcus.Template.WebApi/Startup.cs
```diff
@@ -39,6 +42,20 @@ public void ConfigureServices(IServiceCollection services)
...
+            var openApiInformation = new Info
+            {
+                Title = "Arcus.Template.WebApi",
+                Version = "v1"
+            };
+
+            services.AddSwaggerGen(swaggerGenerationOptions =>
+            {
+                swaggerGenerationOptions.SwaggerDoc("v1", openApiInformation);
+                swaggerGenerationOptions.IncludeXmlComments(Path.Combine(AppContext.BaseDirectory, "Arcus.Template.WebApi.Open-Api.xml"));
```
tomkerkhove: Can we plug in the name of the project that was created based on this template?

src/Arcus.Template.WebApi/Startup.cs
```diff
@@ -39,6 +42,20 @@ public void ConfigureServices(IServiceCollection services)
             });
 
             services.AddHealthChecks();
+
+//#if DEBUG
+            var openApiInformation = new Info
+            {
+                Title = "Arcus.Template.WebApi",
```
tomkerkhove: Can we plug in the name of the project that was created based on this template?

src/Arcus.Template.WebApi/.template.config/template.json
tomkerkhove: I think that's a good suggestion so that they can still uncomment it if they want, what do you think?

src/Arcus.Template.WebApi/Startup.cs
stijnmoreels: Yep, all occurences of the project name (`Arcus.Template.WebApi`) will be replaced by the name you give the created project.

src/Arcus.Template.WebApi/Startup.cs
stijnmoreels: Yep, all occurences of the project name (Arcus.Template.WebApi) will be replaced by the name you give the created project.

src/Arcus.Template.WebApi/Startup.cs
stijnmoreels: Yep, all occurences of the project name (Arcus.Template.WebApi) will be replaced by the name you give the created project.

(Copy-Pasta 🙈)

src/Arcus.Template.WebApi/.template.config/template.json
stijnmoreels: Yeah, it's not so 'clean' but it does the job well and besides from the devs that uses contributes to the project; no else sees the commented directives.

So, if you're do find this a good approach; then OK cool. 👌

src/Arcus.Template.WebApi/Startup.cs
stijnmoreels: image

For example.

src/Arcus.Template.WebApi/Startup.cs
tomkerkhove: Awesome!

src/Arcus.Template.WebApi/Startup.cs
tomkerkhove: Awesome, thanks!

src/Arcus.Template.WebApi/.template.config/template.json
tomkerkhove: I think that's fine, yes, thanks!

src/Arcus.Template.WebApi/Arcus.Template.WebApi.csproj
```diff
@@ -2,6 +2,9 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.2</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DocumentationFile>Arcus.Template.WebApi.Open-Api.xml</DocumentationFile>
+    <!--#if (AuthoringMode)-->
```
tomkerkhove: Is this intentional?

stijnmoreels: Yep, all the tags inside the `AuthoringMode` will be removed when the project is created.

We don't all those tags with Arcus info for example.

tomkerkhove: Perfect, just wanted to make sure :D

PR #25, opened 2019-10-04
FEAT: add NuGet release pipeline

stijnmoreels:
Something like this, the NuGet release pipeline?

Still have to remove the last line from MyGet to NuGet, but was not entirely sure how to test this since we had problems in the past with testing YAML builds with GitHub/NuGet releases from a forked repo.

Relates to #8


build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+            displayName: 'Run integration tests'
+            inputs:
+              command: test
+              projects: 'src/**/$(Project).Tests.Integration.csproj'
+              arguments: '--configuration $(Build.Configuration)'
+              nobuild: true
+              publishTestResults: true
+
+  - stage: Release
+    displayName: 'Release to MyGet'
```
tomkerkhove:
```suggestion
    displayName: 'Release to NuGet.org'
```

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+    dependsOn: IntegrationTests
+    condition: succeeded()
+    jobs:
+      - job: PushToNuGet
+        displayName: 'Push to NuGet'
```
tomkerkhove:
```suggestion
        displayName: 'Push to NuGet.org'
```

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+            inputs:
+              targetPath: '$(Pipeline.Workspace)/templates'
+              artifactName: Templates
+
+  - stage: IntegrationTests
+    displayName: Integration Tests
+    dependsOn: Build
+    condition: succeeded()
+    jobs:
+      - job: RunIntegrationTests
```
tomkerkhove: Thinking out loud and beyond scope of this PR, but would job templates make sense?
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#job-templates

However, I'm wondering if we should run all tests here given we've already tested things, but can't hurt - Right?!

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+              isPreRelease: true
+              compareWith: 'lastRelease'
+            condition: contains(variables['Build.BuildNumber'], '-')
+          - task: NuGetCommand@2
+            displayName: 'Push to NuGet.org'
+            inputs:
+              command: push
+              packagesToPush: 'templates/*.nupkg'
+              nuGetFeedType: external
+              publishFeedCredentials: 'MyGet.org (Push)'
```
tomkerkhove: I've seen your comment but think we can already change this given it works fine so we can ship it!

```suggestion
              publishFeedCredentials: 'NuGet.org'
```

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+          - task: Docker@1
+            displayName: 'Run new web API project Docker image from template'
+            inputs:
+              command: 'Run an image'
+              imageName: 'myproject:$(Build.BuildId)'
+              containerName: 'myproject'
+              ports: '$(Http.Port):$(Http.Port)'
+              envVars: |
+                ARCUS_HTTP_PORT=$(Http.Port)
+          - task: qetza.replacetokens.replacetokens-task.replacetokens@3
```
tomkerkhove: Looks like we can use this YAML template here, right?

https://github.com/arcus-azure/azure-devops-templates/blob/master/test/run-integration-tests.yml#L19

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+    jobs:
+      - job: Compile
+        pool:
+          vmImage: 'ubuntu-16.04'
+        steps:
+          - task: DotNetCoreInstaller@0
+            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.Version))'
+            inputs:
+              version: '$(DotNet.Sdk.Version)'
+          - task: DotNetCoreCLI@2
```
tomkerkhove: Looks like we can use this YAML template here, right?

https://github.com/arcus-azure/azure-devops-templates/blob/master/build/build-solution.yml

build/nuget-release.yml
tomkerkhove: If so, let's create an issue for it

build/nuget-release.yml
```diff
@@ -0,0 +1,172 @@
...
+              gitHubConnection: 'GitHub (arcus-automation - OAuth)'
+              repositoryName: 'arcus-azure/arcus.templates'
+              tagSource: manual
+              tag: 'v$(Build.BuildNumber)'
+              title: 'v$(Build.BuildNumber)'
+              releaseNotesSource: input
+              releaseNotes: |
+                Install new version via [NuGet](https://www.nuget.org/packages/Arcus.Template.WebApi/$(Build.BuildNumber))
+                ```shell
+                > Install-Package Arcus.Template.WebApi -Version $(Build.BuildNumber)
```
stijnmoreels: If the `VersionSuffix` is set (via the YAML template) and not the `Version`; what should this be?

stijnmoreels: Nevermind; the suffix is used to set the version.

build/nuget-release.yml
stijnmoreels: Yeah, job templates could maybe replace some 'basic' steps which are duplicated.

Issue #24, opened 2019-10-02
Testing - Project option test cases


stijnmoreels:
Now, in the integration tests we test the simple possible way of creating a project from the template (without any options).

It would be reasonable if the integration tests consists of several test cases that each creates a new project with one or more options specified and then takes some actions on that new created project to see if the selected options reflect in a changed behavior.

Ex.: when choosing for a authentication mechanism, testing if we indeed have to configure the correct authentication.

tomkerkhove:
That's a good idea, we would then have to add a stage per scenario and run tests for a given category or how would you approach it?

Seperate stages are not a must thought, just easy to discover but requires seperate agents.

stijnmoreels:
I would have the creation of the projects take on a lower level; in the code itself.
That way, we could run it locally and would the YAML build not be bloated with each time a new testing combination.

stijnmoreels:
I'll to come up with an example in a WIP PR.

tomkerkhove:
Sounds good to me, will be easier to grasp with a sample I think since I'm not sure if fully getting your suggestion - Thanks!

Just keep in mind that we're mainly testing all the options so in theory we need to run a `dotnet new` with the option to make sure that the output is what we are testing.

PR #39, opened 2019-10-09
TEST: provide maintainable testing approach for the project options

stijnmoreels:
Relates to #24


src/Arcus.Templates.Tests.Integration/appsettings.local.json
```diff
@@ -1,7 +1,8 @@
 {
   "Arcus": {
     "Api": {
-      "BaseUrl": "http://localhost:5000/api/v1"
+      "BaseUrl": "http://localhost:5000/api/v1",
+      "ProjectPath": "..\\..\\..\\..\\Arcus.Templates.WebApi\\"
```
stijnmoreels: A little help? I need the path to the project/nuget

src/Arcus.Templates.Tests.Integration/Authentication/v1/SharedAccessKeyAuthenticationOptionTests.cs
```diff
@@ -0,0 +1,74 @@
...
+                using (HttpResponseMessage response = await project.Health.GetAsync())
+                {
+                    // Assert
+                    Assert.Equal(HttpStatusCode.Unauthorized, response.StatusCode);
+                }
+            }
+        }
+
+        [Fact]
+        public async Task SharedAccessKeyAuthenticationOption_GetHealthAuthenticated_ResultsOk()
```
stijnmoreels: Example of how you can test shared access key authentication option:
1. provide generated secret
2. create project options model
3. pass along to web project (which will create a temp project)
4. access endpoints on the created project to see the added authentication

src/Arcus.Templates.Tests.Integration/Arcus.Templates.Tests.Integration.csproj
```diff
@@ -24,12 +24,19 @@
...
     <PackageReference Include="Flurl.Http" Version="2.4.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Polly" Version="7.1.1" />
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
   </ItemGroup>
+  
+  <ItemGroup>
+    <Compile Update="Fixture\InMemorySecretProvider.cs">
```
tomkerkhove: Is this on purpose?

stijnmoreels: Yep, that way we can manipulate the to be created project from the template. Can change it a little bit so we don't have this though.

src/Arcus.Templates.Tests.Integration/appsettings.local.json
tomkerkhove: Hm interesting, will have to take a look at this after my holiday. Not sure what the required value is at first glance

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
```diff
@@ -39,11 +49,23 @@ public HealthEndpointService(TestConfig configuration, ITestOutputHelper outputW
         /// </summary>
         public async Task<HttpResponseMessage> GetAsync()
         {
-            _outputWriter.WriteLine("GET {0} ->", _healthEndpoint);
-            var response = await HttpClient.GetAsync(_healthEndpoint);
-            _outputWriter.WriteLine("<- {0}", response.StatusCode);
+            return await GetAsync(request => { });
```
tomkerkhove: What is this request for? Can't we make it optional?

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
```diff
@@ -39,11 +49,23 @@ public HealthEndpointService(TestConfig configuration, ITestOutputHelper outputW
...
+
+        /// <summary>
+        /// Sends a GET request to the health endpoint of the API web application.
+        /// </summary>
+        public async Task<HttpResponseMessage> GetAsync(Action<HttpRequestMessage> alterRequest)
+        {
+            _outputWriter.WriteLine("GET -> {0}", _healthEndpoint);
+            using (var request = new HttpRequestMessage(HttpMethod.Get, _healthEndpoint))
+            {
+                alterRequest(request);
```
tomkerkhove: Maybe better to call a virtual method that those can override in case they have to, rather than passing an action?

src/Arcus.Templates.Tests.Integration/Fixture/TestConfig.cs
```diff
@@ -12,38 +14,89 @@ public class TestConfig : IConfigurationRoot
...
+            return result;
+        }
+
+        private static readonly Random RandomPort = new Random();
+
+        /// <summary>
+        /// Gets the base URL of the to-be-created project from the web API template.
         /// </summary>
-        public string GetBaseUrl()
+        public Uri CreateWebAPiBaseUrl()
```
tomkerkhove:
```suggestion
        public Uri CreateWebApiBaseUrl()
```

src/Arcus.Templates.Tests.Integration/Fixture/TestConfig.cs
```diff
@@ -12,38 +14,89 @@ public class TestConfig : IConfigurationRoot
...
+
+        /// <summary>
+        /// Gets the base URL of the to-be-created project from the web API template.
+        /// </summary>
+        /// <returns></returns>
+        public Uri GetDockerBaseUrl()
+        {
+            const string baseUrlKey = "Arcus:Api:BaseUrl";
+
+            string baseUrl = _configuration.GetValue<string>(baseUrlKey);
```
tomkerkhove: A lot of this method is duplicated in `CreateWebAPiBaseUrl`, can we refactor it out?

src/Arcus.Templates.Tests.Integration/Fixture/InMemorySecretProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    public class InMemorySecretProvider : ICachedSecretProvider
+    {
+        private readonly IDictionary<string, string> _secretValueByName;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="InMemorySecretProvider"/> class.
+        /// </summary>
+        /// <param name="secretValueByName">The sequence of combinations of secret names and values.</param>
+        public InMemorySecretProvider(params (string name, string value)[] secretValueByName)
```
tomkerkhove: You know I'm not a fan of this 😁

Can't we just us a dictionary given you use `ToDictionary` later on anyway?

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        private static async Task WaitUntilWebProjectIsAvailable(int httpPort)
+        {
+            var waitAndRetryForeverAsync =
+                Policy.HandleResult((HttpStatusCode code) => code == HttpStatusCode.NotFound)
+                      .Or<Exception>()
+                      .WaitAndRetryForeverAsync(_ => TimeSpan.FromSeconds(1));
+
+            async Task<HttpStatusCode> GetNonExistingEndpoint()
```
tomkerkhove: Can we move this to a full method please?

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
```diff
@@ -0,0 +1,98 @@
...
+            project.Run(configuration.BuildConfiguration, $"--ARCUS_HTTP_PORT {baseUrl.Port}");
+            await WaitUntilWebProjectIsAvailable(baseUrl.Port);
+
+            return project;
+        }
+
+        private static async Task WaitUntilWebProjectIsAvailable(int httpPort)
+        {
+            var waitAndRetryForeverAsync =
+                Policy.HandleResult((HttpStatusCode code) => code == HttpStatusCode.NotFound)
```
tomkerkhove: Is this enough if the test is enabling Shared access keys?

src/Arcus.Templates.Tests.Integration/Authentication/v1/SharedAccessKeyAuthenticationOptionTests.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Arcus.Templates.Tests.Integration.Fixture;
+using Xunit;
+using Xunit.Abstractions;
+using static Arcus.Templates.Tests.Integration.Fixture.InMemorySecretProvider;
+
+namespace Arcus.Templates.Tests.Integration.Authentication.v1
```
tomkerkhove: Can we add a similar test for Swagger UI?

- Spin up in debug, is able to navigate to Swagger UI
- Spin up in debug, is able to navigate to Swagger docs
- Spin up in release, is not able to navigate to Swagger UI
- Spin up in release, is not able to navigate to Swagger docs

src/Arcus.Templates.Tests.Integration/Fixture/TemplateProject.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            _disposed = true;
+
+            PolicyResult[] results = 
+            {
+                Policy.NoOp().ExecuteAndCapture(() => Disposing(true)),
+                RetryAction(StopProject),
+                RetryAction(UninstallTemplate),
+                RetryAction(DeleteProjectDirectory),
```
tomkerkhove: Idea - Can we make this configurable or so? Reason why I'm asking is that if a test fails we might want to see the project that was generated or am I going too far?

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProjectOptions.cs
```diff
@@ -0,0 +1,80 @@
...
+            Guard.NotNullOrWhitespace(secretValue, nameof(secretValue), "Cannot add shared access key authentication project option without a secret value");
+
+            ProjectOptions optionsWithSharedAccessAuthentication = AddOption(
+                "--Authentication SharedAccessKey",
+                projectDirectory => ConfigureSharedAccessAuthentication(projectDirectory, headerName, secretName, secretValue));
+
+            return new WebApiProjectOptions(optionsWithSharedAccessAuthentication);
+        }
+
+        private void ConfigureSharedAccessAuthentication(DirectoryInfo projectDirectory, string requestHeader, string secretName, string secretValue)
```
tomkerkhove: This looks hacky but probably have to use it? 😁

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Initializes a new instance of the <see cref="HealthDockerEndpointTests"/> class.
+        /// </summary>
+        public HealthDockerEndpointTests(ITestOutputHelper outputWriter)
+        {
+            _configuration = TestConfig.Create();
+            _outputWriter = outputWriter;
+        }
+
+        [Fact]
+        public async Task Health_Get_Docker_Succeeds()
```
tomkerkhove: Thinking out loud - Why is this test different from the non-Docker one and can we make it so that we re-use the same test for different hosting options?

Let's discuss first before changing everything btw 😁

src/Arcus.Templates.Tests.Integration/Authentication/v1/SharedAccessKeyAuthenticationOptionTests.cs
stijnmoreels: Great minds think alike

stijnmoreels: But maybe not in this PR. Could you create an issue for this? Thanks.

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
stijnmoreels: It's actually a kind of fail-safe so we start the test cases only when we have a running instance.
Without knowing exactly what endpoints the API has; we can try to call an enpoint that we could guess doesn't exists and try that until we don't have a connection failure but a `NotFound`.

We could also call the health endpoint in the same way but trying until we get any response of course.

src/Arcus.Templates.Tests.Integration/Fixture/InMemorySecretProvider.cs
stijnmoreels: Yep, will do.

src/Arcus.Templates.Tests.Integration/Arcus.Templates.Tests.Integration.csproj
stijnmoreels: Changed. Not anymore in the project file.

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
stijnmoreels: If we want to manipulate the request before sending. Like in the shared access auth test to add headers.

This overlaod doesn't add anything so we have to pass along an empty action.

src/Arcus.Templates.Tests.Integration/Fixture/TestConfig.cs
stijnmoreels: Yep. Thanks.

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
stijnmoreels: But the test case (ex. shared access key auth) determines how the request should be altered. Therefore an action can be passed allong and the test reader can see in what way the request should be changed. No?

Otherwise we have to create an seperate child class for every altered request. See?

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProjectOptions.cs
stijnmoreels: Yeah, and I guess that for some other features; we also would have to alter the created project in a certain way so it is usable.

This code to add a stub secret provider can be reused though if we want to test the certificate authentication.

src/Arcus.Templates.Tests.Integration/Fixture/TemplateProject.cs
stijnmoreels: It seems that we already have a good logging on the build machine when installing/building/unintalling the template for the tests: https://dev.azure.com/codit/Arcus/_build/results?buildId=38402

Can you explain what you would make configurable?

LIke configure that the project directory is not deleted after the test so we can manually look into it locally?

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
stijnmoreels: Overload to pass along headers instead of action?

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
stijnmoreels: It's different because in the Docker test; we don't have to spin up anything (since the Docker container is running).
I think that we should leave the current YAML build and this test so we know we can run this in a Docker container; but beside this we should have a lot of others to test the project options and combintation of options.

src/Arcus.Templates.Tests.Integration/Fixture/TemplateProject.cs
stijnmoreels: Configurable in a way that the project don't gets deleted so we can manually inspect the folder locally? For example?

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
tomkerkhove: That's a good point, but if we have security enabled, will we ever get NotFound rather than Unauthorized?

src/Arcus.Templates.Tests.Integration/Authentication/v1/SharedAccessKeyAuthenticationOptionTests.cs
tomkerkhove: Done (#43)

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
tomkerkhove: In that case, would it make sense to use another test category for this?

src/Arcus.Templates.Tests.Integration/Health/HealthEndpointService.cs
tomkerkhove: No that is ok, makes sense!

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
stijnmoreels: 1. If you mean using the current approach, we'll always end up with a `NotFound` since the endpoint doesn't exists
2. If you mean using the health, than yes, we'ld have to check if we get _any_ response instead of looking for the right status code.

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
stijnmoreels: I thought that I added a `[Category]` attribute on this class. Yes.

src/Arcus.Templates.Tests.Integration/Fixture/WebApiProject.cs
tomkerkhove: Resolving, looks ok. Health would be nice but too much churn given auth can be different.

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
tomkerkhove: Ah I see now but we have not made any different in the release pipeline.

Would it make sense to run the categories seperately? With that I mean one stage called "Integration Tests" with 2 jobs: 1 for Docker-based tests, 1 for the integration tests without any infra dependency?

stijnmoreels: Yeah, that makes sense. Both the release and CI pipeline then?

tomkerkhove: I'd say yes

tomkerkhove: Added #44 to refactor the builds to reduce duplication

stijnmoreels: Ok! I'll pick this up tomorrow

tomkerkhove: If you want to split it in a new issue that's also fine, but we have time. Just don't want to block the PR on this.

stijnmoreels: Ah, OK; then maybe in a diff issue.
It's not that the tests are not run; so maybe better other issue :).

src/Arcus.Templates.Tests.Integration/appsettings.local.json
stijnmoreels: Is what improved; now we can use the build sources env variable in a CI build; and a simpler path in a local app settings file.

src/Arcus.Templates.Tests.Integration/Fixture/TemplateProject.cs
stijnmoreels: To configure for example to not delete the project dir so we can manually inspect it locally?

src/Arcus.Templates.Tests.Integration/appsettings.local.json
tomkerkhove: 👌

src/Arcus.Templates.Tests.Integration/Fixture/TemplateProject.cs
tomkerkhove: Jup, that is correct!

stijnmoreels: Could help; but I you could also place a breakpoint instead and let the test run afterwards to clean it anyway.

Maybe also something for an issue?

tomkerkhove: Sure, feel free to open one to track this.

Breakpoint certainly works but it could also help troubleshooting issues in Azure DevOps in case of flaky tests.

stijnmoreels: Done #45

src/Arcus.Templates.Tests.Integration/Health/v1/HealthDockerEndpointTests.cs
tomkerkhove: Created #46

stijnmoreels:
> LGTM, only have to fix the 2 remaining comments

I think all is resolved now?

tomkerkhove:
I was not looking at the latest version, sorry 😅

Issue #48, opened 2019-10-17
Provide build error for shared access key generation instead of warnings


tomkerkhove:
As of today we log warnings when we generate a project with Shared Access Key auth here:
- https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.WebApi/Startup.cs#L28
- https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.WebApi/Startup.cs#L41

Given the current state of the API is not valid and should not be released, we should switch this so that it gives build errors rather than warnings so that it's clear to users that they have to change something before they can ship their new API.

What do you think @stijnmoreels ?

stijnmoreels:
I agree that we should indicate to the user that the project needs some additional configuration; but not in the sense that it should not build.
What we could do is change the setup a bit so the project cannot be started correctly.

Also, when we talk about compile errors; that means we also have them when developping the project. Which means we should deactive all these functionality with constants (otherwise we can't build a NuGet package). This could lead to uncomfortable development and a greater feedback loop if something is wrong.

tomkerkhove:
> I agree that we should indicate to the user that the project needs some additional configuration; but not in the sense that it should not build.
> What we could do is change the setup a bit so the project cannot be **started** correctly.

And why are you not a fan of not building it?

Not allowing it to start is too late in the chain in my opinion as the generated template is not in a good-to-go state.

> Also, when we talk about compile errors; that means we also have them when developping the project. Which means we should deactive all these functionality with constants (otherwise we can't build a NuGet package). This could lead to uncomfortable development and a greater feedback loop if something is wrong.

Wouldn't we only have the issue when we run it for shared access key auth?

Personally I don't think this reason should be a blocker and we should work around it.

stijnmoreels:
> Wouldn't we only have the issue when we run it for shared access key auth?
For now because that's our only option; but if diff options about diff contexts gets introduced; could be several.

> And why are you not a fan of not building it?
I think because the additional configuration is about critical parts when the app is running and not when the app is being created/build.

But if you want to change this in build errors, we could do this.

tomkerkhove:
That is correct but from a user experience perspective that approach isn't good because they will only notice it when they are testing their feature which sucks.

I'd personally go for a build error indeed

PR #49, opened 2019-10-18
FIX: force auth config before running

stijnmoreels:
Relates to #48


stijnmoreels:
Hmm, not implemented so quickly; should be discussed first.

PR #55, opened 2019-10-23
FIX: use build errors instead of warnings to incidate missing values

### LINKED ISSUE #48: Provide build error for shared access key generation instead of warnings

stijnmoreels:
Use build errors to indicate user-provided values. In this case, the `secretProvider: null` is removed so we get a compilation error about not finding the right `CachedSecretProvider` constructor.

Closes #48

### DIFF
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -26,7 +26,7 @@ namespace Arcus.Templates.WebApi
         {
 #if Auth
             #warning Please provide a valid secret provider, for example Azure Key Vault: https://security.arcus-azure.net/features/secrets/consume-from-key-vault
-            services.AddScoped<ICachedSecretProvider>(serviceProvider => new CachedSecretProvider(secretProvider: null));
+            services.AddScoped<ICachedSecretProvider>(serviceProvider => new CachedSecretProvider());
 #endif
 
             services.AddMvc(options => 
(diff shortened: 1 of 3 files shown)


src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -26,7 +26,7 @@ public void ConfigureServices(IServiceCollection services)
         {
 #if Auth
             #warning Please provide a valid secret provider, for example Azure Key Vault: https://security.arcus-azure.net/features/secrets/consume-from-key-vault
```
tomkerkhove: Can't we just use `#error` instead?

stijnmoreels: Ah, yes; but we still have to remove the contants to make it buildable for dev. Hang on.

tomkerkhove: That's fine, but I'm more of a fan of the `null` approach as it's up to them to add what they want.

DEV is our issue :D

PR #60, opened 2019-10-25
PATCH: Fix issue with text/plain being used in Swagger UI

### LINKED ISSUE #50: 406 is returned when Accept header is text/plain
406 is returned when the client send the Accept: text/plain

How to fix this?
* remove the text plain as supported media type for JsonOutputFormatter?
```csharp
            var jsonOutputFormatter = options.OutputFormatters.OfType<JsonOutputFormatter>().FirstOrDefault();
            if (jsonOutputFormatter != null)
            {
                if (!jsonOutputFormatter.SupportedMediaTypes.Contains(Text.Plain))
                {
                    jsonOutputFormatter.SupportedMediaTypes.Add(Text.Plain);
                }
            }
```
* remove the text plain as supported media type for JsonInputFormatter? 
```csharp
            var jsonInputFormatter = options.InputFormatters.OfType<JsonInputFormatter>().FirstOrDefault();
            if (jsonInputFormatter != null)
            {
                if (jsonInputFormatter.SupportedMediaTypes.Contains(Text.Plain))
                {
                    jsonInputFormatter.SupportedMediaTypes.Remove(Text.Plain);
                }
            }
```
*  understand why the StringOutputFormatter doesn't kick in.
StringOutputFormatter is added by default and according to the [docs](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/web-api/advanced/formatting.md) it should take care of text/plain
If not used / if doesn't kick in, better to remove
```csharp
            var stringFormatter = options.OutputFormatters.OfType<StringOutputFormatter>().FirstOrDefault();
            if (stringFormatter != null) options.OutputFormatters.RemoveType<StringOutputFormatter>();
```

tomkerkhove:
Signed-off-by: Tom Kerkhove <kerkhove.tom@gmail.com>

image

image

Closes #50

### DIFF
--- a/docs/features/web-api-template.md
+++ b/docs/features/web-api-template.md
@@ -33,7 +33,7 @@ Creates a starter web API project with by default configured:
 ## Configuration
 
 And additional features available with options:
-* `-A|--Authentication` (default `None`)
+* `-au|--authentication` (default `None`)
   * `SharedAccessKey`: adds [shared access key authentication](https://webapi.arcus-azure.net/features/security/auth/shared-access-key) mechanism to the API project
   * `Certificate`: adds [client certificate authentication](https://webapi.arcus-azure.net/features/security/auth/certificate) mechanism to the API project
   * `None`: no authentication configured on the API project.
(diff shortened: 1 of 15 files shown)


MassimoC:
I am not a fan of the Produces attribute because it forces the media type of the error response even if it shouldn't.

I understand it's works fine for a basic scenario. 
BTW, if u want to go this way why adding the attribute on the operation and not on the controller?

tomkerkhove:
@MassimoC it's not that I want this perse but not sure what the alternative is 😁

tomkerkhove:
Another rebase gone wrong, only changed `Startup` & controller.

Should be good to go now

tomkerkhove:
Any feedback on this @MassimoC?

PR #63, opened 2019-11-01
FEAT: use configuration to retrieve certificate subject for certificate authentication

### LINKED ISSUE #14: Provide capability to opt-in for client cert auth during generation
Provide capability to opt-in for client cert auth during generation so that it is already configured.

[.NET Boxed](https://github.com/Dotnet-Boxed/Templates/blob/master/Docs/API.md) is a good example where they allow you to configure what to include or not.

stijnmoreels:
Place the certificate subject name in an `appsettings.json`
configuration file instead of retrieving it via a `ISecretProvider`
implementation (and possible Azure Key Vault).
This is changed because the certificate subject, issuer and thumbprint
can be seen as a publicly known value and not a secret.

Closes #14


src/Arcus.Templates.WebApi/.template.config/template.json
```diff
@@ -87,6 +95,10 @@
     "CertificateAuth": {
       "type": "computed",
       "value": "(authentication == \"Certificate\")"
+    },
+    "AppSettings": {
+      "type": "computed",
+      "value": "(authentication == \"Certificate\")"
```
tomkerkhove: I would make this a seperate option for the template as this can be useful beyond this scope

stijnmoreels: You mean that users are allowed to set if the project should include `appsettings.json` ?

stijnmoreels: I added this because it would make more sense in the source files to use this directive than `CertificateAuth`. But in any case, beyond this scope indeed if you talk about having a seperate option.

Now it's just a compjuted one.

tomkerkhove: It's certainly a good addition but would make it a general one since we don't include it by default, but folks can opt in for it if the want.

Maybe something like `includeAppSettings` or `useAppSettings`?

src/Arcus.Templates.WebApi/.template.config/template.json
```diff
@@ -87,6 +95,10 @@
     "CertificateAuth": {
       "type": "computed",
       "value": "(authentication == \"Certificate\")"
-    } 
+    },
+    "AppSettings": {
+      "type": "computed",
+      "value": "CertificateAuth"
```
tomkerkhove: Not a template expert, but is this correct?

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -0,0 +1,5 @@
+{
```
tomkerkhove: I presume this is the file that is being included if the flag is turned on?

If so, shouldn't we inject this test content since others might not want to have this or not?

src/Arcus.Templates.WebApi/.template.config/template.json
stijnmoreels: This means we have a 'computed' directive that will result in TRUE when the certificate authentication is TRUE.
'Computed' means not configurable for the user but generated at creation-time.

stijnmoreels: So, yes; correct. :)

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: It's not test content; it's required when the certificate authentication is enabled.
We have included a configuration key in the `Startup.cs` that would reflect to this structure.

We could inject both but not sure if that is really something that would make it more user-friendly.

stijnmoreels:
> Sorry for the nitpicking man, just want to get this right from the get go.
> 
> Would you mind updating the docs on the new configuration flag please?

There is not new configuration flag.

stijnmoreels:
Ok, I fixed the build.

src/Arcus.Templates.WebApi/.template.config/template.json
stijnmoreels: Ok, but maybe in other PR; this PR is only about changing the certificate authentication information to the configuration file.

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: Yeah, that was something I would change in an upcoming PR; when we introduce the `--include-appsettings` project options. In the current state; that's the simplest thing for it to work.

src/Arcus.Templates.Tests.Integration/Fixture/CannotStartTemplateProject.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace Arcus.Templates.Tests.Integration.Fixture
+{
+    /// <summary>
+    /// Exception thrown when a project created from a template cannot be started correctly.
+    /// </summary>
+    [Serializable]
+    public class CannotStartTemplateProject : Exception
```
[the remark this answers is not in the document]
stijnmoreels: Woops. Thanks.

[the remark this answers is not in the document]
tomkerkhove: No worries!

src/Arcus.Templates.WebApi/appsettings.json
tomkerkhove: Sounds good to me, tracking in https://github.com/arcus-azure/arcus.templates/issues/64

PR #65, opened 2019-11-06
FEAT: add Open API security definitions as project option

stijnmoreels:
A new project option is introduced to include the security responses to
authorized API endpoints.
This feature required to have an additional controller to be tested
correctly, which means the `WebApiProject` was slightly altered to have
the possibility to add fixture files after the project is created but
before it's started.

### DIFF
--- a/docs/features/web-api-template.md
+++ b/docs/features/web-api-template.md
@@ -37,6 +37,8 @@ And additional features available with options:
   * `SharedAccessKey`: adds [shared access key authentication](https://webapi.arcus-azure.net/features/security/auth/shared-access-key) mechanism to the API project
   * `Certificate`: adds [client certificate authentication](https://webapi.arcus-azure.net/features/security/auth/certificate) mechanism to the API project
   * `None`: no authentication configured on the API project.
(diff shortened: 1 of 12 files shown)


tomkerkhove:
Is this to add [OAuth security definition to API operations](https://webapi.arcus-azure.net/features/openapi/security-definitions)?

If so, I'm not sure if this should be part of the template in my opinion as a very little userbase will need this 🤔 

What do you think @MassimoC?

stijnmoreels:
Hmm, if it's a small userbase, I'm wondering why it's in Arcus web API actually.

tomkerkhove:
For those who need it ;)

The aim for this template is to provide a good starting point with the feature that have to be there for all APIs where in this case it's more for folks who can use this and still pull it in themselves

stijnmoreels:
> For those who need it ;)
> 
> The aim for this template is to provide a good starting point with the feature that have to be there for all APIs where in this case it's more for folks who can use this and still pull it in themselves

Ok, I'll close this PR, then?

tomkerkhove:
I'd suggest to do so if that's ok. Typically we'd first discuss new additions to the template before adding them to avoid this, sorry about that.

PR #69, opened 2019-11-27
Remove legacy for certificate auth & use latest versions of dependencies

### LINKED ISSUE #58: Remove unnecessary middleware method to load client certificate from request header
## Problem
The `Startup` class still has a middleware method to load the client certificate from the `X-ARR-ClientCert` header. This will become unnecessary since the `CertificateAuthenticationFilter` will automatically load this in the future.

## Solution
This filter is part of the [arcus.webapi](https://github.com/arcus-azure/arcus.webapi) repo and will become available for all versions above **v.0.2.0**.

tomkerkhove:
- Remove legacy for certificate auth
- Use latest versions of dependencies, include Arcus Web API

Closes #58

### DIFF
--- a/src/Arcus.Templates.WebApi/Arcus.Templates.WebApi.csproj
+++ b/src/Arcus.Templates.WebApi/Arcus.Templates.WebApi.csproj
@@ -56,7 +56,7 @@
     <PackageReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.2.0" />
-    <PackageReference Include="Arcus.WebApi.Logging" Version="0.2" />
+    <PackageReference Include="Arcus.WebApi.Logging" Version="0.3.0" />
     <PackageReference Include="Arcus.WebApi.Security.Authentication" Version="0.2" Condition="'$(Auth)' == 'true'" />
     <PackageReference Include="Swashbuckle.AspNetCore.SwaggerGen" Version="4.0.1" />
     <PackageReference Include="Swashbuckle.AspNetCore.SwaggerUI" Version="4.0.1" />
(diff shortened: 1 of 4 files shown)


src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -123,46 +120,5 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             });
 //[#endif]
         }
-#if CertificateAuth
```
stijnmoreels: Always fun to remove 🎉

tomkerkhove: Yes it is! It's also fun to open a PR without running the tests so 🤞 😂

tomkerkhove: Too bad, would've been too easy 😂

src/Arcus.Templates.WebApi/Arcus.Templates.WebApi.csproj
```diff
@@ -56,7 +56,7 @@
     <PackageReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.2.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="2.2.0" />
-    <PackageReference Include="Arcus.WebApi.Logging" Version="0.2" />
+    <PackageReference Include="Arcus.WebApi.Logging" Version="0.3.0" />
     <PackageReference Include="Arcus.WebApi.Security.Authentication" Version="0.2" Condition="'$(Auth)' == 'true'" />
```
stijnmoreels: This should also be updated ;) I think here's your problem with the build.

tomkerkhove: Ah it looks like the Package Manager doesn't like those conditions 🙄 
image

PR #70, opened 2019-11-29
CHORE: introduce integration category for Docker & SelfContaining tests

### LINKED ISSUE #46: Enhance our integration tests in builds
Enhance our integration tests in builds so that we have one stage which has 2 jobs:
1. To run the test category for all tests that require the generated image to run as a Docker container
2. To run the test category for integration tests that requires no infrastructure

stijnmoreels:
Uses the integration test YAML template which takes in a category. This allows us to split the 'Docker' integration tests from the 'Self-Containing' integration tests; making our CI build more correctly representing the test targets.

Closes #46

### DIFF
--- a/build/ci-build.yml
+++ b/build/ci-build.yml
@@ -41,39 +41,13 @@ stages:
         pool:
           vmImage: 'ubuntu-16.04'
         steps:
-          - powershell: |
-              $packageVersion = '$(Build.BuildNumber)';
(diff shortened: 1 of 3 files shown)


stijnmoreels:
Used diff stages instead of diff jobs because they can run in parallel.

tomkerkhove:
Awesome, thanks!

Issue #11, opened 2019-09-06
Provide `dotnet new` template for Worker


tomkerkhove:
Provide a [custom `dotnet new`](https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates) template which configures a Worker with minimal requirements.

The template should be calles `arcus-servicebus-queue` and `arcus-servicebus-topic`.

## Features in the template

- [x] TCP Health Probes
- [x] Register `ICachedSecretProvider` as dependency
- [x] Registration of message pump which is using empty message handler
  - By default, use `ISecretProvider` for getting connection string
- [x] Empty message handler

## Good reads

- https://rehansaeed.com/custom-project-templates-using-dotnet-new/

stijnmoreels:
@tomkerkhove , question.
I've come up with a maintainable approach to test the message pump in generated projects from the message pump project template (ServiceBus Queue). But I'm not sure of a few things:

* A message pump should be present in the resulting project without any extra project options?
* Of what other project options do you think? Because if they all relate to the message pump, I'll move some of the functionality at a higher level. Not sure in what direction this template will grow, that's why I ask.

tomkerkhove:
As discussed, the template should:
1. Create a message pump (ie. `OrderMessagePump`) with a temporary message type that logs a message
2. Register the secret provider
3. Register health checks
4. Register message pump with temporary message type

tomkerkhove:
@stijnmoreels This is finished, right?! 🎉

stijnmoreels:
> @stijnmoreels This is finished, right?! 🎉

We've now got a template for the Service Bus Queue; we may need one to for the Service Bus Topic.
Not many work, but I'ld keep this open till that's done. OK?

tomkerkhove:
Sounds good to me!

PR #78, opened 2020-01-22
FEAT: add service bus topic project template

stijnmoreels:
Provides a new project template `arcus-servicebus-topic` to create a project with a message pump on a Azure Service Bus Topic.

Relates to #11

### DIFF
--- a/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
+++ b/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS base
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine3.10 AS base
 WORKDIR /app
 EXPOSE 80
 
--- a/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
+++ b/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
@@ -1,0 +1,19 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS base
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS build
+WORKDIR /src
+COPY ["Arcus.Templates.ServiceBus.Topic.csproj", ""]
+
+COPY . .
+WORKDIR "/src/."
+RUN dotnet build "Arcus.Templates.ServiceBus.Topic.csproj" -c Release -o /app
+
+FROM build AS publish
+RUN dotnet publish "Arcus.Templates.ServiceBus.Topic.csproj" -c Release -o /app
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app .
+ENTRYPOINT ["dotnet", "Arcus.Templates.ServiceBus.Topic.dll"]
--- a/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
+++ b/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS base
 WORKDIR /app
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine3.10 AS build
 WORKDIR /src
 COPY ["Arcus.Templates.ServiceBus.Topic.csproj", ""]
 
--- a/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
+++ b/src/Arcus.Templates.ServiceBus.Topic/Dockerfile
@@ -5,7 +5,6 @@ EXPOSE 80
 FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine3.10 AS build
 WORKDIR /src
 COPY ["Arcus.Templates.ServiceBus.Topic.csproj", ""]
-
 COPY . .
 WORKDIR "/src/."
 RUN dotnet build "Arcus.Templates.ServiceBus.Topic.csproj" -c Release -o /app
--- a/src/Arcus.Templates.ServiceBus.Topic/Program.cs
+++ b/src/Arcus.Templates.ServiceBus.Topic/Program.cs
@@ -1,0 +1,34 @@
+﻿using Arcus.Security.Core;
+using Arcus.Security.Core.Caching;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Arcus.Templates.ServiceBus.Topic
+{
+    public class Program
+    {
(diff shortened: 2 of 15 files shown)


src/Arcus.Templates.Tests.Integration/Worker/ServiceBusWorkerProject.cs
```diff
@@ -19,12 +19,12 @@ namespace Arcus.Templates.Tests.Integration.Worker
     /// <summary>
     /// Project template to create Azure ServiceBus Queue worker projects.
     /// </summary>
-    public class ServiceBusQueueWorkerProject : TemplateProject, IAsyncDisposable
+    public class ServiceBusWorkerProject : TemplateProject, IAsyncDisposable
```
stijnmoreels: @tomkerkhove not sure what would be best. Should I create two child classes or should I place it here in the same class and provide diff methods.

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS base
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS build
+WORKDIR /src
+COPY ["Arcus.Templates.ServiceBus.Topic.csproj", ""]
+
```
tomkerkhove:
```suggestion
```

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS base
```
tomkerkhove: Let's include all security patches
```suggestion
FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine3.10 AS base
```

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+WORKDIR /app
+EXPOSE 80
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.101-alpine3.9 AS build
```
tomkerkhove: Let's include all security patches
```suggestion
FROM mcr.microsoft.com/dotnet/core/sdk:3.0-alpine3.10 AS build
```

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
tomkerkhove: Normally the base image is ASP.Net Core to provide everything related to the runtime but given this is a worker that's overkill.

However, I think we should use `runtime` rather than `sdk`.

tomkerkhove: Not sure if this even works, or does it?

src/Arcus.Templates.ServiceBus.Topic/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+                {
+                    configuration.AddCommandLine(args);
+                    configuration.AddEnvironmentVariables();
+                })
+                .ConfigureServices((hostContext, services) =>
+                {
+                    //#error Please provide a valid secret provider, for example Azure Key Vault: https: //security.arcus-azure.net/features/secrets/consume-from-key-vault
+                    services.AddSingleton<ISecretProvider>(serviceProvider => new CachedSecretProvider(secretProvider: null));
+
+                    services.AddServiceBusTopicMessagePump<EmptyMessagePump>("Receive-All", secretProvider => secretProvider.GetRawSecretAsync("ARCUS_SERVICEBUS_CONNECTIONSTRING"));
```
tomkerkhove: Hm this makes me wonder, don't we have one without having to specify where the connection string comes from and we just give the name of the key? 🤔 

Might be interesting to have a `services.AddServiceBusTopicMessagePump<EmptyMessagePump>("Receive-All", "My-Secret-Name");`, what do you think?

src/Arcus.Templates.Tests.Integration/Worker/ServiceBusWorkerProject.cs
tomkerkhove: This is just for testing so would keep it as is, no?

src/Arcus.Templates.Tests.Integration/appsettings.json
```diff
@@ -10,7 +10,8 @@
         "AuthKey": "#{Arcus.TestInfra.EventGrid.Auth.Key}#"
       },
       "ServiceBus": {
-        "ConnectionStringWithQueue": "#{Arcus.ServiceBus.ConnectionStringWithQueue}#"
+        "ConnectionStringWithQueue": "#{Arcus.ServiceBus.ConnectionStringWithQueue}#",
+        "ConnectionStringWithTopic": "#{Arcus.ServiceBus.ConnectionStringWithTopic}#"
```
tomkerkhove: Did you update the build/variable group?

src/Arcus.Templates.Tests.Integration/Worker/ServiceBusWorkerProject.cs
stijnmoreels: Yeah, testing.
It maybe more 'reasonable' to have sep classes for sep project templates. IDK, that's why I asked 😄 .

src/Arcus.Templates.ServiceBus.Topic/Program.cs
stijnmoreels: Hmm, but then, how do we know it's a configuration key or a (keyvault) secret key?

src/Arcus.Templates.Tests.Integration/appsettings.json
stijnmoreels: This key was already there 😄 probably for the Arcus.Messaging testing.

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
stijnmoreels: Well, all the tests run on the Service Bus resources; so I would say, yes.

src/Arcus.Templates.ServiceBus.Topic/Program.cs
tomkerkhove: It would be for secrets only in that case, discouraging folks to not use configuration.

src/Arcus.Templates.Tests.Integration/appsettings.json
tomkerkhove: Perfect, thanks!

src/Arcus.Templates.ServiceBus.Topic/Program.cs
stijnmoreels: Yeah, makes somewhat sense. But that should be documented thorough.

tomkerkhove: Sure, I'll make an issue for it but good to move forward here.

tomkerkhove: https://github.com/arcus-azure/arcus.messaging/issues/48

stijnmoreels:
Done?

tomkerkhove:
FYI - I've changed the PR to reference the issue since we need docs on it @stijnmoreels

Issue #13, opened 2019-09-06
Provide capability to opt-in for Serilog during generation for Web API


tomkerkhove:
Provide capability to opt-in for Serilog  during generation so that it is already configured.

[.NET Boxed](https://github.com/Dotnet-Boxed/Templates/blob/master/Docs/API.md) is a good example where they allow you to configure what to include or not.

PR #79, opened 2020-01-22
FEAT: add Serilog logging as project option

stijnmoreels:
Provide an project choice option (w/ console logger as default) and an Serilog option to include Serilog infrastructure.

Relates to #13


src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -6,16 +6,49 @@
...
+#endif
 
 namespace Arcus.Templates.WebApi
 {
     public class Program
     {
-        public static void Main(string[] args)
+        public static int Main(string[] args)
         {
+#if Serilog
```
stijnmoreels: This part was inspired and recommended by the docs of Serilog. Not sure if this is fully what we wan.t

tomkerkhove: I would throw in Application Insights, personally, as well

stijnmoreels: Hmm, but then it would not be minimal.
But yeah, I'll add it anyway 😈

tomkerkhove: The template aims at being battle-ready which is a must imo so let's go there indeed!

docs/features/web-api-template.md
```diff
@@ -39,6 +39,9 @@ And additional features available with options:
   * `None`: no authentication configured on the API project.
 * `-ec|--exclude-correlation` (default `false`): excludes the [capability to correlate](https://webapi.arcus-azure.net/features/correlation) between HTTP requests/responses.
 * `-ia|--include-appsettings` (default `false`): includes a `appsettings.json` file to the web API project.
+* `-lo|--logging` (default `Default`)
+  * `Default`: no extra logging mechanism except for the default console logging will be added to the web API project.
```
tomkerkhove: `Default`, `Minimal`, `Console` or `Basic`?

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: Done.

src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -6,19 +6,69 @@
...
-        public static void Main(string[] args)
+#if Serilog
+        private const string ApplicationInsightsInstrumentationKeyName = "ApplicationInsightsInstrumentationKeyName";
+
+#endif
+        public static int Main(string[] args)
         {
+#if Serilog
+            IConfiguration configuration = CreateConfiguration(args);
+            var instrumentationKey = configuration.GetValue<string>(ApplicationInsightsInstrumentationKeyName);
```
tomkerkhove: Do we use a default value or `null`?

tomkerkhove: At least we should add a warning/error

src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -6,19 +6,69 @@
...
+
+            Log.Logger = new LoggerConfiguration()
+                .MinimumLevel.Debug()
+                .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
+                .Enrich.FromLogContext()
+                .WriteTo.Console()
+                .WriteTo.ApplicationInsights(instrumentationKey, new EventTelemetryConverter())
```
tomkerkhove:
```suggestion
                .WriteTo.ApplicationInsights(instrumentationKey, new TraceTelemetryConverter())
```

tomkerkhove: We want to use traces, not events

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -2,4 +2,10 @@
   //#if (CertificateAuth)
   "CertificateSubject": "YOUR CERTIFICATE SUBJECT NAME"
   //#endif
+  //#if (CertificateAuth && Serilog)
+  ,
+  //#endif
+  //#if (Serilog)
```
tomkerkhove: Maybe something like `telemetry:applicationInsights:key` would be more structurized?

docs/features/web-api-template.md
stijnmoreels: Hmm, 'Console' would be better indeed.

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: :+1:

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: Yeah, I'll add a warning

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: It's a shame that I require to make a special condition for that comma 😒 
If we would have a value in the appsettings that is always there; then each if/endif section could include the comma and there would be no problem; but since we haven't that(?) it have to be like this.

Or should I just put in a test value or something? For demostration purposes, but also to help making this conditioning easier?

tomkerkhove: I'm not sure what you are talking about, can you give an example please?

docs/features/web-api-template.md
```diff
@@ -40,6 +40,9 @@ And additional features available with options:
 * `-ia|--include-appsettings` (default `false`): includes a `appsettings.json` file to the web API project.
 * `-ec|--exclude-correlation` (default `false`): excludes the [capability to correlate](https://webapi.arcus-azure.net/features/correlation) between HTTP requests/responses from the API project.
 * `-eo|--exclude-openApi` (default `false`): exclude the [ASP.NET Swagger docs generation and UI](https://docs.microsoft.com/en-us/aspnet/core/tutorials/getting-started-with-swashbuckle?view=aspnetcore-3.1&tabs=visual-studio) from API project.
+* `-lo|--logging` (default `Default`)
+  * `Default`: no extra logging mechanism except for the default console logging will be added to the web API project.
```
tomkerkhove:
```suggestion
  * `Console`: no extra logging mechanism except for the default console logging will be added to the web API project.
```

docs/features/web-api-template.md
```diff
@@ -40,6 +40,9 @@ And additional features available with options:
 * `-ia|--include-appsettings` (default `false`): includes a `appsettings.json` file to the web API project.
 * `-ec|--exclude-correlation` (default `false`): excludes the [capability to correlate](https://webapi.arcus-azure.net/features/correlation) between HTTP requests/responses from the API project.
 * `-eo|--exclude-openApi` (default `false`): exclude the [ASP.NET Swagger docs generation and UI](https://docs.microsoft.com/en-us/aspnet/core/tutorials/getting-started-with-swashbuckle?view=aspnetcore-3.1&tabs=visual-studio) from API project.
+* `-lo|--logging` (default `Default`)
```
tomkerkhove:
```suggestion
* `-lo|--logging` (default `Console`)
```

tomkerkhove:
Did you also verify that everything works manually to ensure that it's wired up correctly?

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: Because we use an empty appsettings, we get the problem when we use project options that require an appsetting. Why? Because of the commas between the values. Only when (in this case) both project options are selected, we require a comma. 
This problem or inconvenience could be fixed if we would add a dummy test value to the appsettings. For demonstration purposes or so. Then each if/endif section in the file could include a trailing comma.

I'll look for a better approach with maybe some replacement logic but other than that sis adding a dummy demonstration value the easiest solution.

tomkerkhove: Oh that's ok then, let's just move on.

stijnmoreels: Ok, done.

stijnmoreels:
> Did you also verify that everything works manually to ensure that it's wired up correctly?

I did prev w/out application insights; but i'll check again.
The tests would also make sure that we have a fully started project when you choose either option; so that's good. But I'll check manually too.

stijnmoreels:
Ok, the traces are written to application insights:
image

tomkerkhove:
Great, thanks!

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -1,5 +1,13 @@
...
+  "CertificateSubject": "YOUR CERTIFICATE SUBJECT NAME",
   //#endif
+  //#if (Serilog)
+  "Telemetry": {
+    "ApplicationInsights": {
+      "InstrumentationKey": "YOUR APPLICATION INSIGHTS APPLICATION KEY"
+    }
+  },
+  //#endif
+  "MyConfigKey":  "MyConfigValue"
```
tomkerkhove: Do we need this?

stijnmoreels: 🤣 that was the thing I explained previously; if we have a setting that we always use, we can use that instead; but otherwise the easiest and most maintainable solution would be a demostration config value (like I said 😄 ).

tomkerkhove: This is funny

PR #81, opened 2020-01-24
CHORE: move integration tests to stage templates

### LINKED ISSUE #44: Investigate if we can use stage/job templates for our test stages
Investigate if we can use stage/job templates for defining our test stages in our builds given they are duplicated in our CI & release build.

- **Stage template:** https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#stage-templates
- **Job template:** https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#job-templates

stijnmoreels:
Closes #44

### DIFF
--- a/build/ci-build.yml
+++ b/build/ci-build.yml
@@ -32,12 +32,6 @@ variables:
     value: '2.2.105'
   - name: 'Project'
     value: 'Arcus.Templates'
-  - name: 'Http.Port'
-    value: '5000'
-  - name: 'Arcus_Api_BaseUrl'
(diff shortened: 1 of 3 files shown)


tomkerkhove:
Just wondering, if we're only using it once - Do we have to move it out?

stijnmoreels:
> Just wondering, if we're only using it once - Do we have to move it out?

No, no, we use it in the `nuget-release.yml` also; but just testing now 😄

PR #90, opened 2020-02-11
FEAT: add JWT authentication project option with a single role

### LINKED ISSUE #89: Feature - add web API project option for JWT role-based authentication
**What feature would you like to have?**
In any app, there may come a time where you want to limit access for diff authenticated users.. Right now, we allow _all_ features of any API by a single shared access key or configured certificate.
I think it would be useful to add also a JWT authentication system with roles for authorization for a more fine-grained control?

**Alternatives you may have considered?**
We could also add the possibility to use different kinds of shared access keys/certificates for different levels of access in the API. Maybe also with roles.

stijnmoreels:
Provide a simple but complete JWT authentication mechanism with a single role as an authentication option in the web API project template. By default is this (like the other authentication mechanisms) added to the entire application, but with the `[Authorize(Roles = "")]` attributes there's room for fine-grained control.

Closes #89


tomkerkhove:
Adding @gverstraete as he's worked with JWT before, do you think it would be good to have this as an opt-in for our Web API template?

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -9,5 +9,10 @@
     }
   },
   //#endif
-  "MyConfigKey":  "MyConfigValue"
+  //#if (JwtAuth)
+  "Jwt": {
+    "Secret": "YOUR SECRET KEY"
```
gverstraete: You validate the user right? why storing the secret (of I assume the user) in the config?

gverstraete: Sorry, it seems to be the signingkey. Then I would rename this property

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -67,8 +88,38 @@ public void ConfigureServices(IServiceCollection services)
...
+                    })
+                    .AddJwtBearer(x =>
+                    {
+                        x.SaveToken = true;
+                        x.TokenValidationParameters = new TokenValidationParameters
+                        {
+                            ValidateIssuerSigningKey = true,
+                            IssuerSigningKey = new SymmetricSecurityKey(key),
+                            ValidateIssuer = true,
+                            ValidIssuer = "entity that generates the token",
```
gverstraete: I think this one should be configurable as well?

tomkerkhove: Agreed

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -67,8 +88,38 @@ public void ConfigureServices(IServiceCollection services)
...
+                            ValidateAudience = true,
+                            ValidAudience = "client of the app"
```
tomkerkhove: Should be configuration as well

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -67,8 +88,38 @@ public void ConfigureServices(IServiceCollection services)
...
+                    new AuthorizationPolicyBuilder()
+                        .RequireRole("Admin")
+                        .RequireAuthenticatedUser()
+                        .Build();
+
+                options.Filters.Add(new AuthorizeFilter(policy));
 #endif
             });
+#if JwtAuth
+            var key = Encoding.UTF8.GetBytes(Configuration["Jwt:SigningKey"]);
```
tomkerkhove: Can't we get that from the ISecretProvider since it's a key?

stijnmoreels: Yeah, that would be cool. The startup isn't async though, not sure if we could do that. Hmm.

tomkerkhove: We should do that, regardless if it's ideal because it's a secret 🤷‍♂

PR #106, opened 2020-02-21
CHORE: update web API projecte template with updated correlation package

### LINKED ISSUE #76: Remove MyGet package sources in web API template
To use the correlation functionality, we needed the MyGet package because no next release was made for the `Arcus.WebApi` library.

When that happens, we should remove the MyGet package source and use the offical NuGet release.

stijnmoreels:
Closes #76


tomkerkhove:
FYI - I've configured renovate on template & event grid proxy which should pick this up for us

PR #151, opened 2020-03-26
Docs - split template features by version

### LINKED ISSUE #147: Document what version introduces a specific feature
Document what version introduces a specific feature so that people know what to expect.

- Only publish documentation for new features when a new version is released
- Provide indication in documentation what version introduces the feature

None

stijnmoreels:
Closes #147

### DIFF
--- a/docs/features/web-api-template.md
+++ b/docs/features/web-api-template.md
@@ -23,16 +23,31 @@ When installed, the template can be created with shortname: `arcus-webapi`:
 ## Features
 
 Creates a starter web API project with by default configured:
+
+![](https://img.shields.io/badge/Available%20starting-v0.1-green?link=https://github.com/arcus-azure/arcus.templates/releases/tag/v0.1.0)
+
 * [Exception middleware](https://webapi.arcus-azure.net/features/logging) to log unhandled exceptions thrown during request processing.
 * Content negotiation that only supports `application/json`.
 * OpenAPI docs generation and UI (only available locally).
+
+### Configuration
+
+And additional features available with options:
+
+* `-A|--Authentication` (default `None`)
+  * `SharedAccessKey`: adds [shared access key authentication](https://webapi.arcus-azure.net/features/
+  * `None`: no authentication configured on the API project.
+
+![](https://img.shields.io/badge/Available%20starting-v0.2-green?link=https://github.com/arcus-azure/arcus.templates/releases/tag/v0.2.0)
+
 * Provides basic health endpoint with [ASP.NET Core health checks](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-2.2) with [OpenAPI support](https://www.codit.eu/blog/documenting-asp-net-core-health-checks-with-openapi/).
 * Docker building file.
 * Default console logger.
 
-## Configuration
+### Configuration
 
 And additional features available with options:
+
 * `-au|--authentication` (default `None`)
   * `SharedAccessKey`: adds [shared access key authentication](https://webapi.arcus-azure.net/features/security/auth/shared-access-key) mechanism to the API project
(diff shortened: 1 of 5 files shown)


docs/features/web-api-template.md
```diff
@@ -23,16 +23,31 @@ When installed, the template can be created with shortname: `arcus-webapi`:
 ## Features
```
tomkerkhove: Actually I would go from:
- Feature 
	- Version 0.1
- Config
	- Version 0.1
- Feature 
	- Version 0.2
- Config
	- Version 0.2

To
- Feature 
	- Version 0.1
	- Version 0.2
- Config
	- Version 0.1
	- Version 0.2

stijnmoreels: But then you have to look at two places to find the available functionality. Now the configuration and security and possibly other headers are sub headers of each version.

tomkerkhove: I don't really agree so let's ask a 3rd person.

Currently have this @SamVanhoutte, what do you think?
image

Or should we use following approach?

Feature
- Version 0.1
- Version 0.2
Config
- Version 0.1
- Version 0.2

tomkerkhove: Any feedback @SamVanhoutte ?

SamVanhoutte: Not really following 100%, but I'd think that you typically are only interested in the version you are using, so why is there no proposal that has 
Version 0.2
- Feature
- Config
Version 0.1
- Feature 
- Config

Might have missed the point, so bear with me.  Also I believe the most recent version should be at the top?

tomkerkhove: That's certainly an option but that assumes you'll duplicate the information from v0.1 in v0.2 then or not?

So what I actually suggest is to move this to the URL :)

`/features/templates/web-api` goes to latest version which is noted at top via badge. At the bottom you can find a list of old docs.
`/v0.1/features/templates/web-api` provides the old docs 

Then the docs are scoped per page and have capability to go back.

Thoughts?

SamVanhoutte: That sounds logical, indeed.  
`As a user`, I just want to see the documentation of the version I'm using and most of the time I don't care about older versions.  Again: up to you two

tomkerkhove: Let's use URLs then!

stijnmoreels: This is actually a problem for every Arcus project.

tomkerkhove: An opportunity you mean ;)

That's correct but nothing we can't fix. Let's start doing that here and if it's good we can roll it out to others

docs/features/web-api-template.md
```diff
@@ -24,21 +24,7 @@ When installed, the template can be created with shortname: `arcus-webapi`:
```
tomkerkhove, line 24: Let's drop a note here on the version we have under `Web API Project Template` or so with a link maybe?

stijnmoreels: You mean specifying which version this is?

tomkerkhove: List all the versions of the documentation such as:

> - Latest
> - [v0.1](/v0.1/features/web-api-template)

stijnmoreels: I did on the bottom a list of the older versions? Isn't that enough?

docs/v0.1/features/web-api-template.md
```diff
@@ -0,0 +1,39 @@
...
+> dotnet new --install Arcus.Templates.WebApi
+```
+
+When installed, the template can be created with shortname: `arcus-webapi`:
+
+```shell
+> dotnet new arcus-webapi --name Arcus.Demo.WebAPI
+```
+
+
```
tomkerkhove, line 22: Let's remove this line
```suggestion
```

docs/features/web-api-template.md
tomkerkhove: LGTM, missed that sorry!

PR #153, opened 2020-03-29
WIP: Feature - warn user during 'dotnet new' about new version

### LINKED ISSUE #149: Determine if we can warn users about new template versions being available
Determine if we can warn users about new template versions being available which they should upgrade to.

Warning in output

None

stijnmoreels:
Closes #149


stijnmoreels:
Maybe not gonna work bc it uses post-actions and requires authorization of the user. Not really user-friendly then if you have to specify each time if you want to check for a more recent update.

tomkerkhove:
Let's abandon it then, nothing we can do

Issue #157, opened 2020-04-02
Use lowercase url's by default in WebApi project


fgheysels:
Would it be an idea that the project that is generated via the `dotnet new arcus-webapi` command specifies by default that lowercase url's are generated ?

This is done by adding this 
```csharp
services.AddRouting(options => options.LowercaseUrls = true);
```

to the `ConfigureServices`  method.

tomkerkhove:
I like the idea, thanks! 

Are you open to contribute it?

fgheysels:
I'm willing to do it.

tomkerkhove:
Done, thanks man!

PR #161, opened 2020-04-02
Specify that endpoints should all be in lowercase

fgheysels:
This is a PR for issue #157

### DIFF
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -83,7 +83,11 @@ namespace Arcus.Templates.WebApi
     
             services.AddScoped(serviceProvider => new CertificateAuthenticationValidator(certificateAuthenticationConfig));
 #endif
-            services.AddRouting(options => options.LowercaseQueryStrings = true);
+            services.AddRouting(options =>
+            {
+                options.LowercaseUrls = true;
+                options.LowercaseQueryStrings = true;
+            });
             services.AddControllers(options => 
             {
                 options.ReturnHttpNotAcceptable = true;
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -50,7 +50,7 @@ namespace Arcus.Templates.WebApi
     public class Startup
     {
 #if Serilog
-        #warning Make sure that the appsettings.json is updated with your Azure Application Insights instrumentation key.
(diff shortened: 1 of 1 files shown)


src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -50,7 +50,7 @@ namespace Arcus.Templates.WebApi
     public class Startup
     {
 #if Serilog
-        #warning Make sure that the appsettings.json is updated with your Azure Application Insights instrumentation key.
+#warning Make sure that the appsettings.json is updated with your Azure Application Insights instrumentation key.
```
fgheysels, line 53: Woops, sorry for re-aligning all those compiler directives; must have something to do with my formatting options in vs.net

fgheysels: Fixed this. :)

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -83,6 +83,7 @@ public void ConfigureServices(IServiceCollection services)
     
             services.AddScoped(serviceProvider => new CertificateAuthenticationValidator(certificateAuthenticationConfig));
 #endif
+            services.AddRouting(options => options.LowercaseQueryStrings = true);
```
tomkerkhove, line 86: I presume you mean `LowercaseUrls` instead of `LowercaseQueryStrings`?

If `LowercaseQueryStrings` uses camel casing I'd be happy to add that one as well

fgheysels: OMG, haste makes waste.

I've added the LowercaseUrls settings and kept the LowercaseQueryStrings as well.  (The latter one can only be applied if the former one is set to true as well).

tomkerkhove: What does a query parameter look like then - Is it `fooBar` or `foobar`?

fgheysels: I would expect it to be `foobar` , but I've just tested it and it is `fooBar` (my query parameter is declared like that as well), so I'm confused now.

tomkerkhove: lol that's odd. Do you want to keep it in then or not?

fgheysels: I think we should keep it in

https://github.com/dotnet/aspnetcore/issues/10746

PR #166, opened 2020-04-07
Feature - add OpenApi security definitions for authentication mechanisms

### LINKED ISSUE #154: Document generated security mechanism in OpenAPI
Document generated security mechanism in OpenAPI when applicable.

`AddSecurityDefinition` should be used when configuring OpenAPI docs for API key authentication.

https://github.com/domaindrivendev/Swashbuckle.AspNetCore#add-security-definitions-and-requirements

**Additional context**

stijnmoreels:
Closes #154

### DIFF
--- a/src/Arcus.Templates.Tests.Integration/WebApi/WebApiProjectOptions.cs
+++ b/src/Arcus.Templates.Tests.Integration/WebApi/WebApiProjectOptions.cs
@@ -221,7 +221,7 @@ namespace Arcus.Templates.Tests.Integration.WebApi
 
         private static string InsertSharedAccessAuthenticationHeaderSecretPair(string startupContent, string requestHeader, string secretName)
         {
-            return startupContent.Replace("YOUR REQUEST HEADER NAME", requestHeader)
+            return startupContent.Replace("X-Api-Key", requestHeader)
                                  .Replace("YOUR SECRET NAME", secretName);
         }
 
--- a/src/Arcus.Templates.Tests.Integration/WebApi/WebApiProjectOptions.cs
+++ b/src/Arcus.Templates.Tests.Integration/WebApi/WebApiProjectOptions.cs
@@ -221,7 +221,7 @@ namespace Arcus.Templates.Tests.Integration.WebApi
 
         private static string InsertSharedAccessAuthenticationHeaderSecretPair(string startupContent, string requestHeader, string secretName)
         {
-            return startupContent.Replace("X-Api-Key", requestHeader)
+            return startupContent.Replace("X-API-Key", requestHeader)
                                  .Replace("YOUR SECRET NAME", secretName);
         }
 
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -185,7 +185,7 @@ namespace Arcus.Templates.WebApi
                     { 
                         new OpenApiSecurityScheme
                         {
-                            Description = "Globally authentication scheme based on shared access key"
+                            Description = "Globally authentication scheme based on shared access key",
                             Reference = new OpenApiReference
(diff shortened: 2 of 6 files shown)


stijnmoreels:
All manually tested via the Swagger UI.

src/Arcus.Templates.Tests.Integration/WebApi/WebApiProjectOptions.cs
```diff
@@ -221,7 +221,7 @@ private static string InsertInMemorySecretProviderCode(string startupContent, st
 
         private static string InsertSharedAccessAuthenticationHeaderSecretPair(string startupContent, string requestHeader, string secretName)
         {
-            return startupContent.Replace("YOUR REQUEST HEADER NAME", requestHeader)
+            return startupContent.Replace("X-Api-Key", requestHeader)
```
stijnmoreels, line 224: Changed this because it may be a bit more user-friendly on start-up and you want to discover the API via the Swagger UI.

tomkerkhove: Fine for me - Let's change it to `X-API-Key` then!

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -52,6 +53,10 @@ namespace Arcus.Templates.WebApi
 {
     public class Startup
     {
+#if SharedAccessKeyAuth
+        private const string SharedAccessKeyHeaderName = "X-Api-Key";
```
tomkerkhove, line 57: This is actually a good reminder as we still have to document this header as a request header for SAS Auth. Can you add this here or a dedicated PR please?

tomkerkhove:
LGTM but is there a way that we can test this automatically @stijnmoreels ?

src/Arcus.Templates.WebApi/Startup.cs
stijnmoreels: Hmm not entirely sure what you mean by that 🤔 .

tomkerkhove: It's not documented in OpenAPI similar to this https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.WebApi/Startup.cs#L166

stijnmoreels:
> LGTM but is there a way that we can test this automatically @stijnmoreels ?

Hmm, maybe with Selenium we could run through the Swagger UI and check this.

tomkerkhove:
I think it's good enough if we just parse the OpenAPI spec that is generated, no?

stijnmoreels:
It depends on what you want to test. Is it the parsing into json or the functionality that it describes.

src/Arcus.Templates.WebApi/Startup.cs
stijnmoreels: Oh, but isn't this done with the security definition? Like you have to "Authorize" the request in the Swagger UI and there you can fill out the header or query that is needed. No?

tomkerkhove:
I just want to assert that it's in the JSON spec for different auth options, don't really care about the UI as it's based on that as well

src/Arcus.Templates.WebApi/Startup.cs
tomkerkhove: That's correct but also a header that should be listed as request header/query parameter

tomkerkhove: Anyhow, it's ok

stijnmoreels: But we could still add it?😬

tomkerkhove: Nah it's ok

Issue #138, opened 2020-03-23
Use enrichers from Arcus Observability


tomkerkhove:
Use enrichers from Arcus Observability

stijnmoreels:
Which ones?
- [ ] `VersionEnricher`
- [ ] `ApplicationEnricher`
- [ ] `CorrelationInfoEnricher`
- [ ] `KubernetesEnricher`

tomkerkhove:
- Always include `VersionEnricher`
- Leaning towards always including for `ApplicationEnricher` with decent sample value, such as "API" & "Service Bus Worker"
- Include `CorrelationInfoEnricher` when using correlation (existing flag)
- Not sure about Kubernetes enricher, might be opt-in by using `--kubernetes-ready` flag or so

stijnmoreels:
> Not sure about Kubernetes enricher, might be opt-in by using --kubernetes-ready flag or so

Maybe to specific and "too easy to add" to make it as a project option?

tomkerkhove:
Yes, let's leave out for now

PR #175, opened 2020-04-21
Feature - use Arcus.Observability Serilog enrichers

stijnmoreels:
Relates to #138


docs/preview/servicebus-queue-worker-template.md
```diff
@@ -0,0 +1,36 @@
+---
```
stijnmoreels, line 1: Added a new `./docs/preview` place to set upcoming releases instead of changing the ones in `./docs/features`. This will be easier to manage both new features while still providing a place for the most recent published features.

tomkerkhove: Can't we just add the next version already in the docs or add "Next release" in the badge name so that we have one file?

Just thinking out loud.

src/Arcus.Templates.ServiceBus.Queue/Program.cs
```diff
@@ -87,6 +87,8 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                 .MinimumLevel.Debug()
                 .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
                 .Enrich.FromLogContext()
+                .Enrich.WithVersion()
```
tomkerkhove, line 90: We should add correlation as well or is this pending a release?

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -319,6 +319,8 @@ private LoggerConfiguration CreateLoggerConfiguration(IServiceProvider servicePr
                 .MinimumLevel.Debug()
                 .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
                 .Enrich.FromLogContext()
+                .Enrich.WithVersion()
```
tomkerkhove, line 322: We should add correlation as well or is this pending a release?

src/Arcus.Templates.ServiceBus.Topic/Program.cs
```diff
@@ -87,6 +87,8 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                 .MinimumLevel.Debug()
                 .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
                 .Enrich.FromLogContext()
+                .Enrich.WithVersion()
```
tomkerkhove, line 90: We should add correlation as well or is this pending a release?

src/Arcus.Templates.ServiceBus.Queue/Program.cs
stijnmoreels: No, correlation in message pumps are on message-level so are not configured here. They are still not yet released though, so waiting for the next version of Arcus.Messaging.
See this line: https://github.com/arcus-azure/arcus.messaging/blob/37703b956c6f950bcec6e658a3bdf2b1f28e71a9/src/Arcus.Messaging.Pumps.ServiceBus/AzureServiceBusMessagePump.cs#L450

src/Arcus.Templates.WebApi/Startup.cs
stijnmoreels: See above.

stijnmoreels: It is added below 😄 . See the line `.WithHttpCorrelationInfo`.

src/Arcus.Templates.ServiceBus.Topic/Program.cs
stijnmoreels: See above, same awnser as with the Queue worker.

docs/preview/servicebus-queue-worker-template.md
stijnmoreels: Yes, but if we than actually release the "next version' we have to split it either way. And then we have to look into what has changed and not and all.

tomkerkhove: Yes but then it's in the same file at least, now the diff just says there is a new file :)

stijnmoreels:
So... I can merge, @tomkerkhove ?

docs/preview/servicebus-queue-worker-template.md
tomkerkhove: And why is the reason why we can't move it to the current files?

stijnmoreels: Yes, I think there is maybe a little miscommunication here.
The reason why I created a whole new file, is that if version v0.3 is released, we just have to move the current `.\features\**` files to `.\v0.3\features\**` and move the `.\preview\features\**` to the "most recent published  features".
It allows us to fully change how the docs of this new version should look like. If we would place the changes in the same file, we would have possible duplicates of everything. Every title, every section, every code example... because there could be diff's between the already released version and the version we're working on. Moving them in sepearate files solves that problem.

stijnmoreels: By duplicating it in the same file, you make it a lot harder for devs and consumers to look for the docs they need. Moving it to a seperate file solves that too.

tomkerkhove: Ok, now I'v got it; sorry!

Issue #186, opened 2020-05-06
Introduce Secret Store concept


tomkerkhove:
Introduce Secret Store concept that we've added in #69

PR #197, opened 2020-06-03
Feature - introduce secret store concept into templates

stijnmoreels:
Updates the web API template and both worker templates with using the secret stores instead of registering the `ISecretProvider` by themselves.
Plus add the `IConfiguration` to the stores when running in DEBUG.

Relates to #186

### DIFF
--- a/src/Arcus.Templates.ServiceBus.Queue/Program.cs
+++ b/src/Arcus.Templates.ServiceBus.Queue/Program.cs
@@ -65,7 +65,7 @@ namespace Arcus.Templates.ServiceBus.Queue
                        {
 //[#if DEBUG]
                            stores.AddConfiguration(config);
-                           //[#endif]
+//[#endif]
 
                            //#error Please provide a valid secret provider, for example Azure Key Vault: https: //security.arcus-azure.net/features/secrets/consume-from-key-vault
                            stores.AddAzureKeyVaultWithManagedServiceIdentity("https://your-keyvault-vault.azure.net/");
--- a/src/Arcus.Templates.ServiceBus.Queue/Program.cs
+++ b/src/Arcus.Templates.ServiceBus.Queue/Program.cs
@@ -63,8 +63,12 @@ namespace Arcus.Templates.ServiceBus.Queue
                        })
                        .ConfigureSecretStore((config, stores) =>
                        {
+//[#if DEBUG]
+                           stores.AddConfiguration(config);
+//[#endif]
+
                            //#error Please provide a valid secret provider, for example Azure Key Vault: https: //security.arcus-azure.net/features/secrets/consume-from-key-vault
-                           stores.AddProvider(secretProvider: null);
+                           // stores.AddAzureKeyVault(secretProvider: null);
                        })
 #if (ExcludeSerilog == false)
                        .UseSerilog(UpdateLoggerConfiguration)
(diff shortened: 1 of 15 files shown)


src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -76,6 +76,17 @@ private static IHostBuilder CreateHostBuilder(string[] args, IConfiguration conf
...
                     .ConfigureAppConfiguration(configBuilder => configBuilder.AddConfiguration(configuration))
+#if (SharedAccessKeyAuth || JwtAuth)
+                    .ConfigureSecretStore((config, stores) =>
+                    {
+//[#if DEBUG]
+                        stores.AddConfiguration(config);
+//[#endif]
+
+                        #error Please provide a valid secret provider, for example Azure Key Vault: https://security.arcus-azure.net/features/secrets/consume-from-key-vault
+                        // stores.AddAzureKeyVault(secretProvider: null);
```
tomkerkhove, line 87: I would uncomment this and use the overload which requires the vault name and use MSI by default

src/Arcus.Templates.ServiceBus.Topic/Program.cs
```diff
@@ -61,14 +61,20 @@ public static IHostBuilder CreateHostBuilder(string[] args)
...
                            configuration.AddEnvironmentVariables();
                        })
+                       .ConfigureSecretStore((config, stores) =>
+                       {
+//[#if DEBUG]
+                           stores.AddConfiguration(config);
+//[#endif]
+
+                           //#error Please provide a valid secret provider, for example Azure Key Vault: https: //security.arcus-azure.net/features/secrets/consume-from-key-vault
+                           // stores.AddAzureKeyVault(secretProvider: null);
```
tomkerkhove, line 71: I would uncomment this and use the overload which requires the vault name and use MSI by default

src/Arcus.Templates.ServiceBus.Queue/Program.cs
```diff
@@ -61,14 +61,20 @@ public static IHostBuilder CreateHostBuilder(string[] args)
...
                            configuration.AddEnvironmentVariables();
                        })
+                       .ConfigureSecretStore((config, stores) =>
+                       {
+//[#if DEBUG]
+                           stores.AddConfiguration(config);
+//[#endif]
+
+                           //#error Please provide a valid secret provider, for example Azure Key Vault: https: //security.arcus-azure.net/features/secrets/consume-from-key-vault
+                           // stores.AddAzureKeyVault(secretProvider: null);
```
tomkerkhove, line 71: I would uncomment this and use the overload which requires the vault name and use MSI by default

stijnmoreels:
> Can we remove the secret providers from startup or don't we have those anymore?

What do you mean exactly? 😅

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: I added the NuGet package that contains this extension and left it there without any conditions in the csproj file. Therefore, it will be included by default (even without any authentication options).

src/Arcus.Templates.ServiceBus.Topic/Program.cs
stijnmoreels: I added the NuGet package that contains this extension and left it there without any conditions in the csproj file. Therefore, it will be included by default (even without any authentication options).

src/Arcus.Templates.ServiceBus.Queue/Program.cs
stijnmoreels: I added the NuGet package that contains this extension and left it there without any conditions in the csproj file. Therefore, it will be included by default (even without any authentication options).

PR #222, opened 2020-08-25
Use Microsoft.NET.Sdk.Worker project type for workers

tomkerkhove:


stijnmoreels (approved):
Oh, wow, cool. Didn't know that. Thanks!

Issue #244, opened 2020-10-15
Provide template for Azure Service Bus worker in Azure Functions


tomkerkhove:
**Is your feature request related to a problem? Please describe.**
Provide a template for Azure Service Bus worker in Azure Functions which uses message handler concept with all our best practices baked into it.

**Additional context**
Depends on https://github.com/arcus-azure/arcus.messaging/issues/127

stijnmoreels:
Both for queue and topic? Or is an Azure Function able to switch easily between the two.

tomkerkhove:
It is, ask me again in a week and I'll be able to give you a sample (repo is private for now)

stijnmoreels:
Waiting for https://github.com/arcus-azure/arcus.messaging/issues/127 to be complete.

stijnmoreels:
Waiting for https://github.com/arcus-azure/arcus.templates/pull/265 to be complete since it has some Azure Functions test changes that we can benefit from.
Already have some work done locally, though, with the MyGet package. Looking good!

stijnmoreels:
Waiting for Arcus.Messaging as it's not .NET 6 supported.

PR #270, opened 2021-01-08
🚧 Feature - add project template for Azure Functions Service Bus Queue

stijnmoreels:
Add project template for the Azure Functions Service Bus Queue that uses the Arcus Messaging router to handle incoming messages on the Azure Service Bus Queue.

Relates to #244


stijnmoreels:
Waiting till we have at least a preview version of the Arcus Messaging library that includes the extraction of message routing from the message pumps. https://github.com/arcus-azure/arcus.messaging/pull/141

PR #303, opened 2021-02-19
Feature - small JSON serialization improvements

stijnmoreels:
Some JSON serialization settings that were present in the Azure Functions HTTP trigger, but not in the web API.

### DIFF
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -94,7 +94,7 @@ namespace Arcus.Templates.WebApi
                 options.RespectBrowserAcceptHeader = true;
 
                 RestrictToJsonContentType(options);
-                AddEnumAsStringRepresentation(options);
+                ConfigureJsonFormatters(options);
 
 #if SharedAccessKeyAuth
                 #warning Please provide a valid request header name and secret name to the shared access filter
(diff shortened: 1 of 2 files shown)


src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -247,17 +247,21 @@ private static void RestrictToJsonContentType(MvcOptions options)
             options.OutputFormatters.RemoveType<StringOutputFormatter>();
         }
 
-        private static void AddEnumAsStringRepresentation(MvcOptions options)
+        private static void ConfigureJsonFormatters(MvcOptions options)
         {
             var onlyJsonInputFormatters = options.InputFormatters.OfType<SystemTextJsonInputFormatter>();
             foreach (SystemTextJsonInputFormatter inputFormatter in onlyJsonInputFormatters)
             {
+                inputFormatter.SerializerOptions.MaxDepth = 10;
```
tomkerkhove, line 255: Should we do this?

stijnmoreels: I'm more of the opinion to have this in place and that the developer change it to something higher if they need it, then to leave it open to 64 by default and maybe overload something.

tomkerkhove: WDYT @fgheysels ?

fgheysels: In my humble opinion, we should stick to the defaults defined in .NET.
Target audience is developers, so we shouldn't be too restrictive or 'parental' (or whatever you call it :) ).

tomkerkhove: Frankly, I was thinking the same on this one

stijnmoreels: Ok, then should we also remove it in the Azure Function HTTP trigger 'cause it's defined there too. I can do that in this PR too.

stijnmoreels: Left the `IgnoreNullValues` still there, but removed the `MaxDepth` in both the web API as the Azure Functions HTTP trigger template. Resolving.

tomkerkhove (approved):

PR #359, opened 2021-04-27
fix: set Microsoft log level to warning

### LINKED ISSUE #354: Reduce verbosity for telemetry by Microsoft
Telemetry should be actionable and not generate noise, by default Microsoft telemetry is added which is considered to be noise.

Reduce verbosity for telemetry by Microsoft by reducing it to Warning-level only.

https://github.com/arcus-azure/arcus.templates/blob/master/src/Arcus.Templates.WebApi/Startup.cs#L314

N/A

**Additional context**
This generates a lot of logging statements when executing an HTTP endpoint.

stijnmoreels:
To reduce logs from Microsoft, we should set it to warning.

Closes #354

### DIFF
--- a/src/Arcus.Templates.WebApi/Startup.cs
+++ b/src/Arcus.Templates.WebApi/Startup.cs
@@ -311,7 +311,7 @@ namespace Arcus.Templates.WebApi
             
             return new LoggerConfiguration()
                 .MinimumLevel.Debug()
-                .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
+                .MinimumLevel.Override("Microsoft", LogEventLevel.Warning)
                 .Enrich.FromLogContext()
                 .Enrich.WithVersion()
                 .Enrich.WithComponentName("API")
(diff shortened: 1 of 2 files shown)


tomkerkhove (approved):

fgheysels (approved):
Thanks !

src/Arcus.Templates.WebApi/Startup.cs
```diff
@@ -311,7 +311,7 @@ private LoggerConfiguration CreateLoggerConfiguration(IServiceProvider servicePr
             
             return new LoggerConfiguration()
                 .MinimumLevel.Debug()
```
fgheysels, line 313: Is it a good idea to set the loglevel to Debug ?  I mean, Arcus is a library, customers who use this library in their project might get overwhelmed with Arcus logs that they do not want.

stijnmoreels: Or set it on Debug when the configuration is Debug 😄 .

tomkerkhove: +1. I'd use info by default

fgheysels: I'd set it to Info, regardless if the configuration is debug or release.

PR #372, opened 2021-05-06
fix: use information from config as minimum Serilog log level

### LINKED ISSUE #354: Reduce verbosity for telemetry by Microsoft

stijnmoreels:
Use the configuration specified in the `appsettings.json` of the web API project template to specify that the minimum Serilog log level should be `Information`.

Closes #354 
Relates to https://github.com/arcus-azure/arcus.templates/discussions/368
Relates to https://github.com/arcus-azure/arcus.templates/discussions/355

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -229,7 +229,7 @@ GEM
     rb-fsevent (0.10.4)
     rb-inotify (0.10.1)
       ffi (~> 1.0)
-    rexml (3.2.4)
+    rexml (3.2.5)
     rouge (3.23.0)
     ruby-enum (0.8.0)
       i18n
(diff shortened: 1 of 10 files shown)


tomkerkhove (approved):

tomkerkhove:
Do we set configuration for Docker as well? Otherwise we need to add that env var, but OK if it's left out

stijnmoreels:
> Do we set configuration for Docker as well? Otherwise we need to add that env var, but OK if it's left out

Not sure to what Docker setting you're referring to, so I guess we don't set it. 😅

tomkerkhove (approved):

PR #395, opened 2021-07-27
fix: Resolve a NU1608 warning in the Arcus.Templates.Tests.Integration.csproj

sucrose0413:
Hi, I found a NU1608 warning in the Arcus.Templates.Tests.Integration.csproj:

`Warning NU1608 Detected package version outside of dependency constraint: Microsoft.Azure.ServiceBus 4.1.1 requires System.IdentityModel.Tokens.Jwt (>= 5.4.0 && < 6.0.0) but version System.IdentityModel.Tokens.Jwt 6.11.1 was resolved.`

This fix can remove the warning from Arcus.Templates's dependency graph.
Hope the PR can help you.

Best regards,
sucrose

### DIFF
--- a/src/Arcus.Templates.Tests.Integration/Arcus.Templates.Tests.Integration.csproj
+++ b/src/Arcus.Templates.Tests.Integration/Arcus.Templates.Tests.Integration.csproj
@@ -52,6 +52,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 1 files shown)


stijnmoreels:
Hi @sucrose0413 ! Thanks for your PR!
The reason there is a warning, is because a conflict in our packages (see `Arcus.WebApi` <> `Arcus.Messaging`). The messaging package is still using the older version of the Azure Service Bus SDK (`Microsoft.Azure.ServiceBus`) but that will change when we upgrade the message library (soon). Then, this conflict will probably be resolved by itself.

stijnmoreels (approved):

PR #475, opened 2021-11-19
feat: update web api proj template to net6.0

### LINKED ISSUE #472: Migrate `Arcus.Templates.WebApi` towards .NET 6
We should migrate the `Arcus.Templates.WebApi` project towards .NET 6 and use the new hosting features and setup.

**Additional context**
https://docs.microsoft.com/en-us/aspnet/core/migration/50-to-60-samples?view=aspnetcore-6.0
Relates to https://github.com/arcus-azure/arcus/issues/106

stijnmoreels:
Updates the Web API project template towards .NET 6.

Closes #472 
Relates to https://github.com/arcus-azure/arcus/issues/106


stijnmoreels:
- [ ] Updates CI .NET SDK version

build/ci-build.yml
```diff
@@ -59,6 +59,11 @@ stages:
               keepToken: false
               tokenPrefix: '#{'
               tokenSuffix: '}#'
+          - task: UseDotNet@2
+            displayName: 'Import .NET Core SDK ($(DotNet.Sdk.PreviousVersion))'
+            inputs:
+              packageType: 'sdk'
+              version: '$(DotNet.Sdk.PreviousVersion)'
```
fgheysels, line 66: Why do we need to import the previous version of the sdk ?

src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -22,21 +53,24 @@ public class Program
...
         private const string ApplicationInsightsInstrumentationKeyName = "Telemetry:ApplicationInsights:InstrumentationKey";
-
+        
+#endif
+#if SharedAccessKeyAuth
+        private const string SharedAccessKeyHeaderName = "X-API-Key";
+        
 #endif
         public static int Main(string[] args)
         {
```
fgheysels, line 63: Should we migrate here to the same minimized version of Program.cs like the one that is generated by VS.NET 2022 when starting a new project ?
(One with top level statements, so no namespace declaration, no class specification, no Main method .... ) ?

It might become too complex though.

src/Arcus.Templates.WebApi/Program.cs
```diff
@@ -131,5 +335,68 @@ private static IHostBuilder CreateHostBuilder(string[] args, IConfiguration conf
                 .WriteTo.AzureApplicationInsights(instrumentationKey);
         }
 #endif
+
```
fgheysels, line 338: Just a suggestion; these methods are coming from the Startup class, but we do not have a `Startup` any more in .NET 6.
Is it OK to have these methods directly in this file, or should we consider creating a partial class so that we can put these things in a separate file ?
Just thinking out loud.

build/ci-build.yml
stijnmoreels: Because the other project templates are not yet updated, and still uses this .NET Core version.

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: Hmm, we should be careful of introducing new syntactic sugar to our common templates. Might be a good idea to stay with something that's common ground.
People can always change it.

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: That would be a good idea for maintaining the project. But maybe people do want to maintain this differently, like adding extensions for related functionality. Hmm 🤔, I like the idea but maybe we shouldn't force this on people. Just want to be as comfortable as possible for everyone 😅.

Maybe this can be checked later, some experimentation is maybe in order. 🤔

fgheysels: That's a bit the 'problem' with a template, you're always 'forced' in a certain way of working.

src/Arcus.Templates.WebApi/Program.cs
fgheysels: Well, it's just to be more aligned with how the VS.NET project templates for web apps look like.
I think this new syntax can be good for simple things; in our case, we do already have a lot of extra methods etc....   I agree to keep it this way for now.

build/ci-build.yml
fgheysels: Ah, ok.  And this build-ci is used for all templates.

fgheysels (approved):
I think this is OK.
Great that this has been picked up already

src/Arcus.Templates.WebApi/Program.cs
stijnmoreels: We can still move to the new syntax later, when people are maybe more accustomed to it.

build/ci-build.yml
stijnmoreels: Yep indeed 😉

Issue #484, opened 2021-12-03
Use `System.Text.Json` instead of `Newtonsoft.Json` for JSON (de)serialization 


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
We still use the separate package `Newtonsoft.Json` for JSON (de)serialization. `System.Text.Json` is by-default present in the .NET environment.

**Describe the solution you'd like**
We should consider switching to the Microsoft-package `System.Text.Json` as default.

**Additional context**
Discussed in https://github.com/arcus-azure/arcus/discussions/171

stijnmoreels:
Waiting for 'missing members' support: https://github.com/dotnet/runtime/issues/37483

fgheysels:
> Waiting for 'missing members' support: [dotnet/runtime#37483](https://github.com/dotnet/runtime/issues/37483)

Looks like that issue is fixed.
Apparently, it helps when somebody puts a bit of [anger ](https://github.com/dotnet/runtime/issues/37483#issuecomment-1318350596) in his voice.

stijnmoreels:
Hehe, and he's got a point. 😄

PR #485, opened 2021-12-03
chore: move az func http trigger to system.text.json

stijnmoreels:
Use `System.Text.Json` in Azure Functions HTTP trigger project template

Relates to #484


src/Arcus.Templates.AzureFunctions.Http/HttpBasedAzureFunction.cs
```diff
@@ -1,209 +1,191 @@
...
+using System.Threading.Tasks;
+using Arcus.WebApi.Logging.Correlation;
+using GuardNet;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Formatters;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Net.Http.Headers;
+
```
fgheysels, line 17: Is it just me, or is this diff ui not working as expected ?

fgheysels (approved):

src/Arcus.Templates.AzureFunctions.Http/HttpBasedAzureFunction.cs
stijnmoreels: Yeah, something about new line chars that are probably not right. I fixed it locally, so it won't happen again.

PR #489, opened 2021-12-09
feat: upgrade az func http trigger to net6.0

### LINKED ISSUE #488: Migrate `Arcus.Templates.AzureFunctions.Http` towards .NET 6
**Describe the solution you'd like*
We should migrate Arcus.Templates.AzureFunctions towards .NET 6.

Additional context
https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v4%2Cwindows%2Ccsharp%2Cportal%2Cbash%2Ckeda
Relates to arcus-azure/arcus#106

stijnmoreels:
Upgrades the Azure Functions HTTP Trigger project template to .NET 6.

Closes #488

### DIFF
--- a/src/Arcus.Templates.AzureFunctions.Http/Dockerfile
+++ b/src/Arcus.Templates.AzureFunctions.Http/Dockerfile
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/azure-functions/dotnet:4 AS base
 WORKDIR /app
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/core/sdk:6.0.100-buster AS build
+FROM mcr.microsoft.com/dotnet/core/sdk:6.0.100-bullseye-slim AS build
 WORKDIR /src
 COPY ["Arcus.Templates.AzureFunctions.Http.csproj", ""]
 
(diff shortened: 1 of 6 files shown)


src/Arcus.Templates.AzureFunctions.Http/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/azure-functions/dotnet:3.0 AS base
+FROM mcr.microsoft.com/azure-functions/dotnet:4 AS base
 WORKDIR /app
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1-buster AS build
+FROM mcr.microsoft.com/dotnet/sdk:6.0.100-bullseye-slim AS build
```
fgheysels, line 5: What is our strategy for selecting base images ? 
Buster & bullseye are both images for debian Linux.  I have no problem with that, but I just want to understand why we're using these images here, and for webapi for instance, we're using alpine.

stijnmoreels: I think it's because of the compatibility with Azure Functions. I don't think we can use `alpine` here.

fgheysels: I thought it would be something like that; unfortunately, I was not able to find any documentation on that.

fgheysels (approved):

PR #495, opened 2021-12-16
feat: Azure Functions template should use configurable Serilog settings

### LINKED ISSUE #480: Azure Function templates should  be updated to use configurable serilog settings
The templates that are used to quickly start an Azure Function project should be modified so that  the Serilog configuration can be read from the configuration using `LoggerConfiguration.ReadFrom.Configuration()` just like we do in the Api template.

Next to that, we should also only log to AppInsights if the instrumentationkey is not blank.

spasal:
Implements configurable Serilog settings for Azure Function templates

Closes #480

### DIFF
--- a/src/Arcus.Templates.AzureFunctions.Http/Arcus.Templates.AzureFunctions.Http.csproj
+++ b/src/Arcus.Templates.AzureFunctions.Http/Arcus.Templates.AzureFunctions.Http.csproj
@@ -66,6 +66,7 @@
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.OpenApi" Version="1.0.0" Condition="'$(ExcludeOpenApi)' == 'false'" />
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.14.0" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="3.1.0" />
+    <PackageReference Include="Serilog.Settings.Configuration" Version="3.3.0" />
     <PackageReference Include="Serilog.Sinks.Console" Version="4.0.1" />
   </ItemGroup>
 </Project>
--- a/src/Arcus.Templates.AzureFunctions.Http/Startup.cs
+++ b/src/Arcus.Templates.AzureFunctions.Http/Startup.cs
@@ -52,21 +52,27 @@ namespace Arcus.Templates.AzureFunctions.Http
                 stores.AddAzureKeyVaultWithManagedIdentity("https://your-keyvault.vault.azure.net/", CacheConfiguration.Default);
             });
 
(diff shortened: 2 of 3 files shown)


spasal:
Hi @stijnmoreels @fgheysels ,

A few questions:
 - Should the integration tests be updated to check whether the config settings are really used in the serilog implementation?
 - Any suggestions on how I can test this change locally? Create a project from the local template?

I noticed that there seems to be something wrong with the application insights custom logging levels. From what I saw, you need to specify it as well in the serilog settings of the config as well as in the application insights section of the config in host.json. Should this be included in this PR or should I open a separate issue / PR to investigate this further?

Thanks for the guidance/feedback/help!

stijnmoreels:
Great to have you working on this! This repo is not an easy one to contribute, so awesome job! 👍 

To test that the Serilog actually uses the local file configuration, we could create a test that disrupts the configuration files after the test project is being created, and see that the project's endpoints aren't available.

This can be done with the `UpdateFileInProject` method of the project instance. (see web API project template integration tests for including app settings, for example).

stijnmoreels:
For the other issue about logging levels: we could indeed look into that more after we have established a baseline for reading the configuration in Serilog.
To be clear on this: you mean that logging to Azure Application Insights only works with logging levels when both the Serilog and Application Insights-specific sections in the `host.json` file are updated?

stijnmoreels:
Let us know if you come across any problems. We'll hold off with our review until this PR is out of construction 😉 .

spasal:
> For the other issue about logging levels: we could indeed look into that more after we have established a baseline for reading the configuration in Serilog. To be clear on this: you mean that logging to Azure Application Insights only works with logging levels when both the Serilog and Application Insights-specific sections in the `host.json` file are updated?

Indeed, that seemed to be the case at one of my projects. If you only specify the logLevels in the Serilog section of the `host.json` file, more was logged than specified. Only when I also put the logLevels in the applicationInsights logging section in the `host.json` file, the set logLevels were respected. Note that this was a function that also had some other issues, so will replicate to another function 

Will first focus on the goal for this PR, and after it we can look at the potential issue/fix for the custom loggingLevels

spasal:
> Let us know if you come across any problems. We'll hold off with our review until this PR is out of construction 😉 .
> 
> Ps.: Oh, and a good tip while writing tests: the project instance has `TearDownOptions` which is great to see how the template creates the project. `TearDownOptions.KeepProjectDirectory` will for example keep the generated project after the test is run so you can see on your local disk how the project was created.

One of the things that would make testing locally easier is if someone could provide me their appsettings.private.json, could that be arranged? :)

stijnmoreels:
> One of the things that would make testing locally easier is if someone could provide me their appsettings.private.json, could that be arranged? :)

I think you only need the instrumentation key of Application Insights, but sure, I can provide it. For your future contributions 😉

fgheysels:
Hi @spasal , is this something you still want to keep working on ?

Issue #547, opened 2022-03-31
Use correct configuration-value for configuring ApplicationInsights integration


fgheysels:
When you create a new Azure Function based on the Arcus FunctionApp template, some boilerplate code is generated for setting up logging to Application Insights.
This [code](https://github.com/arcus-azure/arcus.templates/blob/370ab3b65b5caebfd74f9bffc86d51b02ac3153b/src/Arcus.Templates.AzureFunctions.Http/Startup.cs#L55) looks in the configuration to find the instrumentation-key of Application Insights in a setting called `APPLICATIONINSIGHTS_INSTRUMENTATIONKEY`.

However, I've found out that when you use a setting other then `APPINSIGHTS_INSTRUMENTATIONKEY`, Function invocations (Host.Result logs) are not being logged.   They do not end up in the `requests` table in Application Insights, which means that those Function invocations do not appear in the predefined 'Failures', 'Performance', etc... dashboards in Application Insights.

Therefore, we should use `APPINSIGHTS_INSTRUMENTATIONKEY` as the config-setting that contains ApplicationInsights' telemetry-key.

PR #550, opened 2022-04-05
feat: add az func az servicebus queue project template

stijnmoreels:
Adds an Azure Functions project template with an Azure Service Bus trigger on a queue, using the most-recent Arcus packages.

Relates to #244
Relates to #547


src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/OrdersAzureServiceBusMessageHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+using Arcus.Templates.AzureFunctions.ServiceBus.Queue.Model;
+using GuardNet;
+using Microsoft.Extensions.Logging;
+
+namespace Arcus.Templates.AzureFunctions.ServiceBus.Queue
+{
+    /// <summary>
+    /// Represents an <see cref="IAzureServiceBusMessageHandler{TMessage}"/> implementation that processes <see cref="Order"/> messages.
+    /// </summary>
+    public class OrdersAzureServiceBusMessageHandler : IAzureServiceBusMessageHandler<Order>
```
stijnmoreels, line 16: Other than the worker template, I chose to use an 'order' in our project template instead of the `EmptyMessage` variant currently in the worker template. Seemed like a better and more understandable way when first opening the project.

stijnmoreels:
Let me add a Docker test too.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN dotnet build "Arcus.Templates.AzureFunctions.ServiceBus.Queue.csproj" -c Release -o /app
+
+FROM build AS publish
+RUN dotnet publish "Arcus.Templates.AzureFunctions.ServiceBus.Queue.csproj" -c Release -o /app
+
+FROM base AS final
+WORKDIR /app
+COPY --from=publish /app .
+ENV AzureWebJobsScriptRoot=/app \
+    AzureFunctionsJobHost__Logging__Console__IsEnabled=true
```
fgheysels, line 20: Just a question, is this a custom environment variable/setting ?  Or is this a documented variable for Azure Functions ?

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/OrdersAzureServiceBusMessageHandler.cs
fgheysels: Agree.  It will be more straightforward to understand I think.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Startup.cs
```diff
@@ -0,0 +1,66 @@
...
+            IConfiguration appConfig = builder.GetContext().Configuration;
+            var instrumentationKey = appConfig.GetValue<string>("APPINSIGHTS_INSTRUMENTATIONKEY");
+            var logConfig = new LoggerConfiguration()
+                .MinimumLevel.Debug()
+                .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
+                .Enrich.FromLogContext()
+                .Enrich.WithComponentName("Service Bus Queue Trigger")
+                .Enrich.WithVersion()
+                .WriteTo.Console()
+                .WriteTo.AzureApplicationInsights(instrumentationKey);
```
fgheysels, line 61: I always tend to only add the Application Insights sink if the instrumentation-key is provided. (not empty).
It's debatable :)

src/Arcus.Templates.Tests.Integration/Worker/Fixture/TestOrderAzureServiceBusMessageHandler.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
```
fgheysels, line 1: What exactly are we testing here ?
You're testing if the generated project based on the template works as expected ? 

I was more thinking: maybe we should test if the generated files are as expected?  But this might actually be a good approach as well.

fgheysels (approved):
Great work!

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Dockerfile
stijnmoreels: This is an built-in environment variable that was already part of the Dockerfile when it got generated. So, I guess we can leave it. 😄

src/Arcus.Templates.Tests.Integration/Worker/Fixture/TestOrderAzureServiceBusMessageHandler.cs
stijnmoreels: Yeah, the integration tests in the templates library all test the actual workings. So, if a setting of a project option results in an extra file, it checks the general workings instead of just that file. Seems a good approach to test the resulted project as a whole, no? So, in this case, we test indeed if we can receive an Azure Service Bus message via our generated project from the template. That way, we're sure that the template works.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Startup.cs
stijnmoreels: Yes, I saw that of you 😉 . I see if the other templates have this as well, if so, I'm happy to add it here. 👍

stijnmoreels: Hmm, I can't find it in other templates, but I wouldn't mind if we do add this to all our templates. I do, however would like this to be logged when the instrumentation key is blank, because it could lead to unexpected results if we 'magically' skip the Azure Application Insights Serilog sink from being added.

stijnmoreels: What is the benefit from checking first the instrumentation key? That the project don't gets stopped because of a missing logging configuration?

fgheysels: If you do not have this setting configured, the project will crash and fail to start.
You can look at it in 2 ways:
- either you leave it as is, and your program crashes.   Then let the developer decide if he wants to remove the sink to Application Insights, or needs to add the required setting
- Only add the application insights sink if the instrumention-key is provided.  If it is not provided, log a warning.   I've used this approach in one of my projects because I had situations where my application needed to log to App Insights, but when deployed on prem (on a vessel), no app-insights logging should happen.

src/Arcus.Templates.Tests.Integration/Worker/Fixture/TestOrderAzureServiceBusMessageHandler.cs
fgheysels: I agree.  This is the best approach.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Dockerfile
fgheysels: Sure. :) It just made me wonder, because I once ran into a situation where I had dan issue with environment-variables with double underscores, and then I heard from MSFT that it was not advised to use this in Azure Functions, but I have no idea about the underlying reason.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Startup.cs
stijnmoreels: Ok, great! Then let's add this to all our templates and make sure our user-experience is even better! 😉 
https://github.com/arcus-azure/arcus.templates/issues/563

stijnmoreels:
Azure Databricks cluster is gone. So, our Job Metrics tests fail now.

fgheysels:
> Azure Databricks cluster is gone. So, our Job Metrics tests fail now. 

So, this means we can only complete this PR once the databricks cluster is up again ?  Who is going to take care of that ?

stijnmoreels:
> > Azure Databricks cluster is gone. So, our Job Metrics tests fail now.
> So, this means we can only complete this PR once the databricks cluster is up again ? Who is going to take care of that ?

Depends what's the outcome is of the discussion, I guess. If we are allowed to setup such a cluster again, or if we can use the AI Incubator cluster, maybe... Depends.

stijnmoreels:
We can also skip it for now, the tests, and come back on it later. Maybe that's the best way here, otherwise all PR's will be blocked by this.

PR #553, opened 2022-04-08
fix: use http-specific correlation in az func http trigger

stijnmoreels:
Since we updated the Azure Functions extensions to use HTTP-specific correlation options instead, we should update the registration in our Azure Functions HTTP trigger.


stijnmoreels:
Let's not do this.

PR #558, opened 2022-04-13
fix: use correct appinsights instrumentation key env var

### LINKED ISSUE #528: Remove `.ClearProvidersExceptFunctionProviders` from Azure Functions project templates
As discussed in this issue https://github.com/arcus-azure/arcus.observability/issues/336, we should no longer use the `.ClearProvidersExceptFunctionProviders` extension in Azure Functions projects.

Remove the `.ClearProvidersExceptFunctionProviders` call in all the Azure Functions project templates.

**Additional context**
https://github.com/arcus-azure/arcus.observability/issues/336

### LINKED ISSUE #547: Use correct configuration-value for configuring ApplicationInsights integration

stijnmoreels:
Use 'APPINSIGHTS_INSTRUMENTATIONKEY' as environment variable to set instrumentation key.

Closes #547
Closes #528

### DIFF
--- a/src/Arcus.Templates.AzureFunctions.Databricks.JobMetrics/Startup.cs
+++ b/src/Arcus.Templates.AzureFunctions.Databricks.JobMetrics/Startup.cs
@@ -39,7 +39,7 @@ namespace Arcus.Templates.AzureFunctions.Databricks.JobMetrics
                 stores.AddAzureKeyVaultWithManagedIdentity("https://your-keyvault.vault.azure.net/", CacheConfiguration.Default);
             });
 
-            var instrumentationKey = config.GetValue<string>("APPLICATIONINSIGHTS_INSTRUMENTATIONKEY");
+            var instrumentationKey = config.GetValue<string>("APPINSIGHTS_INSTRUMENTATIONKEY");
             var configuration = new LoggerConfiguration()
                 .MinimumLevel.Debug()
                 .MinimumLevel.Override("Microsoft", LogEventLevel.Information)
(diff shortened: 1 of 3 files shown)


src/Arcus.Templates.AzureFunctions.Databricks.JobMetrics/Startup.cs
```diff
@@ -39,7 +39,7 @@ public override void Configure(IFunctionsHostBuilder builder)
                 stores.AddAzureKeyVaultWithManagedIdentity("https://your-keyvault.vault.azure.net/", CacheConfiguration.Default);
             });
 
-            var instrumentationKey = config.GetValue<string>("APPLICATIONINSIGHTS_INSTRUMENTATIONKEY");
+            var instrumentationKey = config.GetValue<string>("APPINSIGHTS_INSTRUMENTATIONKEY");
```
fgheysels, line 42: Just wondering if we have settings file that need to be changed as well ?

stijnmoreels: In this case, no, bc this is set via the integration test project, and this was already done in this PR.
So, it's like this:

Azure DevOps library >-sets-> Integration Test project >-sets-> Project template.

fgheysels (approved):

PR #559, opened 2022-04-13
chore: treat warnings as errors

stijnmoreels:
Treat warnings as errors in the project template projects.

Relates to https://github.com/arcus-azure/arcus/issues/196

### DIFF
--- a/src/Arcus.Templates.AzureFunctions.Http/Startup.cs
+++ b/src/Arcus.Templates.AzureFunctions.Http/Startup.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using Arcus.Security.Core.Caching.Configuration;
 using Arcus.Templates.AzureFunctions.Http;
+using Arcus.WebApi.Logging.Core.Correlation;
 using Microsoft.Azure.Functions.Extensions.DependencyInjection;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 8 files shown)


fgheysels (approved):

src/Arcus.Templates.AzureFunctions.Http/Startup.cs
```diff
@@ -35,7 +36,7 @@ public override void Configure(IFunctionsHostBuilder builder)
         {
             IConfiguration config = builder.GetContext().Configuration;
 
-            builder.AddHttpCorrelation();
+            builder.AddHttpCorrelation((HttpCorrelationInfoOptions options) => { });
```
fgheysels, line 39: This change is because the parameterless `AddHttpCorrelation` method is made obsolete.
However, if you want to add http correlation with just the default options, then I think it is a bit sad that you now need to call this method like this.  (With an empty lambda).  IMHO, it pollutes your code a bit.
Therefore, why don't we keep the parameterless method as an overload, and behind the scenes, call  `AddHttpCorrelation( HttpCorrelationOptions options => {} )`?

fgheysels: This discussion is outside the scope of this PR, but it caught my attention now that I see this here.

stijnmoreels: Yes, this change is added bc this PR adds 'treat warnings as errors' and since this deprecated call was a warning, it was an error in the build 😄 . I know, I would want to use the default too, but we can't do that until the next Web API major.

stijnmoreels: We could make this a 'no warn' in the 'author' section of the template project file, so that it can still use the empty lambda but doesn't fail on the deprecated call. Maybe that's an idea.

stijnmoreels: Done in https://github.com/arcus-azure/arcus.templates/pull/559/commits/09ba884f7fdef747fc9028a72da1466d0bdcbbc1

PR #572, opened 2022-04-27
feat: add az func az servicebus topic template

### LINKED ISSUE #244: Provide template for Azure Service Bus worker in Azure Functions

### LINKED ISSUE #562: Provide template for Azure Service Bus worker consuming a Topic in Azure Functions
It is very similar to [this](https://github.com/arcus-azure/arcus.templates/issues/244) request.
I think we should also provide a project template for a Function that consumes messages that are published to a topic, using a given subscriber.

stijnmoreels:
Adds an Azure Functions Azure Service Bus topic project template.

Closes #562 
Closes #244


src/Arcus.Templates.AzureFunctions.ServiceBus.Topic/OrderFunction.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// Process an Azure Service Bus topic <paramref name="message"/> as an <see cref="Order"/>.
+        /// </summary>
+        /// <param name="message">The incoming message on the Azure Service Bus topic, representing an <see cref="Order"/>.</param>
+        /// <param name="log">The logger instance to write informational messages during the message processing.</param>
+        /// <param name="cancellationToken">The token to cancel the message processing.</param>
+        [FunctionName("order-processing")]
+        public async Task Run(
+            [ServiceBusTrigger("order-topic", "order-subscription",  Connection = "ServiceBusConnectionString")] ServiceBusReceivedMessage message,
+            ILogger log,
```
fgheysels, line 46: Should we create a template that receives one message by default, or multiple messages ?

stijnmoreels: Not sure if it really matters here? What will usually be used? Batches? We should be careful indeed as to not force too much on people.

fgheysels: Indeed , it all depends on the use case.
However, creating a function that receives multiple messages at once, is simply done by specifing that the input object is an array instead of a single message.  We can leave it as it is imho.

fgheysels (approved):

Issue #491, opened 2021-12-10
HealthReport types cause that additional (unwanted) types end up in the OpenApi definition


fgheysels:
When we create an API project using the Arcus template, we get a Health endpoint for free, which is a good thing.

However, the models that are used by this endpoint to create the response have a nasty side-effect:  The Open API specification for our API suddenly includes schema-information for types that consumers of the API are actually not interested in.  In other words, our OpenAPI documentation gets polluted.
It looks like types from System.Reflection are included:

![image](https://user-images.githubusercontent.com/3605786/145550182-48e68444-c55a-4806-95ef-1e5e0c5a3d93.png)

It's easy to work around this, by removing the `ProducedResponseType` attributes from the health-endpoint, however, then the OpenAPI documentation is not complete.

I wonder if we should use a custom type for the response-model of the health-endpoint, as it is clear the the `HealthReport` types cause this pollution.

stijnmoreels:
Hmm, I'm even wondering if we use the `HealthReport` type 🤔. I think we even use `string` as general response type.

stijnmoreels:
I think we even use `string` in Azure Functions bc `HealthReport` gave errors 🤔.

fgheysels:
This is what the health endpoint returns
```json
{
  "entries": {},
  "status": "Healthy",
  "totalDuration": "00:00:00.0000078"
}
```

stijnmoreels:
Maybe schema filters is an option here too?

stijnmoreels:
The `HealthReport` has an `Exception` has `HealthReportEntry` property; I think this leads to this unwanted tree of related types.
A dedicated JSON data transfer object is maybe a good idea. This is something we could consider for the other project templates too.

fgheysels:
Could indeed be that the Exception property causes this, but should we even include exception information in the response of that health-endpoint request ?
I'd say we create our own HealthReport API model for this, which does not include exception info.

stijnmoreels:
Yes, that would be better, indeed. We may even opt to expose this separately in `Arcus.WebApi` so we can both use it in our templates as well as in the TCP probe in Messaging.

PR #589, opened 2022-05-20
feat: add healthreport json to webapi template

stijnmoreels:
Adds a custom `HealthReportJson` JSON data-transfer object to the WebAPI project template with the necessary feature documentation.

Relates to https://github.com/arcus-azure/arcus.templates/issues/491

### DIFF
--- a/docs/preview/features/web-api-template.md
+++ b/docs/preview/features/web-api-template.md
@@ -49,3 +49,11 @@ As part of this template the following HTTP header(s) are removed for security s
 * `Server` header * Provides information concerning the Web API runtime
 
 The OpenAPI documentation is available by-default. Be careful of exposing sensitive information with the OpenAPI documentation, only expose what's necessary and hide everything else.
+
+### Health
+
+A default health controller is available that exposes the configured health checks as an aggregated health report. 
+For more information on application health, see [Microsoft's documentation](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks).
+
+The controller doesn't directly exposes Microsoft's `HealthReport` model but uses a custom `HealthReportJson` operation model which eliminates the exception details from the original report.
+This way the application's health can be exposed in a safe manner without also exposing exception and assembly information to the user.
--- a/src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
+++ b/src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using GuardNet;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
+{
+    /// <summary>
+    /// Represents an alternative <see cref="HealthReport"/> model without the exception details so it can be exposed safely with OpenAPI.
+    /// </summary>
+    /// <remarks>
+    ///     This model should not be used within the domain and is just available for the serialization between trust boundaries.
+    ///     Use the <see cref="FromHealthReport"/> and <see cref="ToHealthReport"/> to switch between the two.
+    /// </remarks>
+    public class HealthReportJson
+    {
+        /// <summary>
+        /// Gets a dictionary containing the results from each health check.
+        /// </summary>
+        public IDictionary<string, HealthReportEntryJson> Entries { get; set; }
+
+        /// <summary>
+        /// Gets a <see cref="HealthStatus"/> representing the aggregate status of all the health checks.
+        /// The value of <see cref="Status"/> will be the most severe status reported by a health check.
+        /// If no checks were executed, the value is always <see cref="HealthStatus.Healthy"/>.
+        /// </summary>
+        public HealthStatus Status { get; set; } = HealthStatus.Healthy;
+
+        /// <summary>
+        /// Gets the time the health check service took to execute.
+        /// </summary>
+        public TimeSpan TotalDuration { get; set; }
+
+        /// <summary>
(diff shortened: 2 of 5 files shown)


stijnmoreels:
A proposal how this would look for Web API.

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
```diff
@@ -0,0 +1,77 @@
...
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
+{
+    /// <summary>
+    /// Represents an alternative <see cref="HealthReport"/> model without the exception details so it can be exposed safely with OpenAPI.
+    /// </summary>
+    /// <remarks>
+    ///     This model should not be used within the domain and is just available for the serialization between trust boundaries.
+    ///     Use the <see cref="FromHealthReport"/> and <see cref="ToHealthReport"/> to switch between the two.
+    /// </remarks>
+    public class HealthReportJson
```
fgheysels, line 17: Why do we have `Json`in the name ?  Suppose we use XML serialization, the model will be returned as Xml, so I would not include 'Json' in the name here.

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using GuardNet;
+
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.Diagnostics.HealthChecks
```
fgheysels, line 8: IMO, the class in this namespace is an API model.   Although it is not a 'business API model' but more an infrastructural model, I think I'd put it in a 'Models' namespace.

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
fgheysels: Ah, I see .... You want to avoid name-clashes.
Maybe `HealthStatusReport` ?

stijnmoreels: Yeah, it's for the name clashes 😅 . Since our template is based on JSON serialization and we even have MVC extensions that only allows JSON serialization to the template, it seemed logically to also focus on JSON in the API model. If it were XML that needs to be serialized, then the startup functionality needs to be changed regardless and the `[JsonProperty]` stuff too. I kinda like it that it reflects in the name what it supposed to do, but we can change it indeed. `HealthStatusReport` will maybe confuse people, though, it would confuse me in any case 😅 .

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
stijnmoreels: Now it's in the serialization namespace as it's for serialization, a model namespace would be more for a business model, no? But, ok, we can move it there.

stijnmoreels: Done in https://github.com/arcus-azure/arcus.templates/pull/589/commits/7da2befc3a9356df33cd00e8a891b83ff54b70d6

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
fgheysels: I just meant that imho it is not a good idea to include such an 'implementation detail' in the name of a type.

`HealthStatusReport`, `HealthResponse`, `HealthReportResponse`, `ApiHealthReport`, `HealthReportResult`, `HealthResult`, `HealthInformation` , .... are imho better then adding the Json suffix :)

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
fgheysels: Though decision :) 
I'd consider that types in a `Serialization` namespace are more 'infrastructural' types that you need to perform serialization.

src/Arcus.Templates.WebApi/Serialization/HealthReportJson.cs
stijnmoreels: `ApiHealthReport` seems like a good one, then.

docs/preview/features/web-api-template.md
```diff
@@ -55,4 +55,12 @@ And additional features available with options:
...
-The OpenAPI documentation is available by-default. Be careful of exposing sensitive information with the OpenAPI documentation, only expose what's necessary and hide everything else.
\ No newline at end of file
+The OpenAPI documentation is available by-default. Be careful of exposing sensitive information with the OpenAPI documentation, only expose what's necessary and hide everything else.
+
+### Health
+
+A default health controller is available that exposes the configured health checks as an aggregated health report. 
+For more information on application health, see [Microsoft's documentation](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks).
+
+The controller doesn't directly exposes Microsoft's `HealthReport` model but uses a custom `HealthReportJson` operation model which eliminates the exception details from the original report.
```
fgheysels, line 65:
```suggestion
The controller doesn't directly exposes Microsoft's `HealthReport` model but uses a custom `ApiHealthReport`  model which eliminates the exception details from the original report.
```

fgheysels (approved):
Just a remark regarding documentation that still refered to the original type-name , but otherwise, this looks good :)

docs/preview/features/web-api-template.md
stijnmoreels: Ah, yes, good catch!

PR #613, opened 2022-06-23
feat: add empty APPINSIGHTS_INSTRUMENTATIONKEY to appsettings.json

### LINKED ISSUE #602: WebApi template improvement: add APPINSIGHTS_INSTRUMENTATIONKEY to appsettings.json
When you create a new WebApi using the Arcus template, code is generated to enable application insights logging.
In the code, you can find this:
image

But the appsettings.json file that is included in the project, doesn't include a property for `APPINSIGHTS_INSTRUMENTATIONKEY`.  I think it would be good if we would add an empty property to the appsettings.json file.

vtermaat:
WebApi template improvement: add APPINSIGHTS_INSTRUMENTATIONKEY to appsettings.json
Closes #602

### DIFF
--- a/src/Arcus.Templates.WebApi/appsettings.json
+++ b/src/Arcus.Templates.WebApi/appsettings.json
@@ -5,7 +5,8 @@
   //#if (Serilog)
   "Serilog": {
     "MinimumLevel": {
-      "Default": "Information"
+      "Default": "Information",
+      "APPINSIGHTS_INSTRUMENTATIONKEY": ""
     }
   },
   //#endif
@@ -15,6 +16,5 @@
     "Audience": "<your-audience>"
   },
   //#endif
-  "MyConfigKey": "MyConfigValue",
-  "APPINSIGHTS_INSTRUMENTATIONKEY": ""
+  "MyConfigKey": "MyConfigValue"
 }
--- a/src/Arcus.Templates.WebApi/appsettings.json
+++ b/src/Arcus.Templates.WebApi/appsettings.json
@@ -15,5 +15,6 @@
     "Audience": "<your-audience>"
   },
   //#endif
-  "MyConfigKey": "MyConfigValue"
+  "MyConfigKey": "MyConfigValue",
+  "APPINSIGHTS_INSTRUMENTATIONKEY": ""
 }
--- a/src/Arcus.Templates.WebApi/appsettings.json
+++ b/src/Arcus.Templates.WebApi/appsettings.json
@@ -3,10 +3,10 @@
   "CertificateSubject": "<your-certificate-subject-name>",
   //#endif
   //#if (Serilog)
+  "APPINSIGHTS_INSTRUMENTATIONKEY": "",
   "Serilog": {
     "MinimumLevel": {
-      "Default": "Information",
-      "APPINSIGHTS_INSTRUMENTATIONKEY": ""
+      "Default": "Information"
     }
   },
   //#endif


stijnmoreels:
We will need to make sure that this is only being added when the Serilog is used and not Console.

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -15,5 +15,6 @@
     "Audience": "<your-audience>"
   },
   //#endif
-  "MyConfigKey": "MyConfigValue"
+  "MyConfigKey": "MyConfigValue",
+  "APPINSIGHTS_INSTRUMENTATIONKEY": ""
```
stijnmoreels, line 19: Will have to be moved within the `//#if (Serilog)` as the Application Insights is only used when the project is configured using Serilog.

stijnmoreels (changes requested):
Let's move this within the right condition, and we're good to go! 👍 

Ps.: also thinking about other templates that can benefit from this. Will need to check.

src/Arcus.Templates.WebApi/appsettings.json
vtermaat: https://github.com/arcus-azure/arcus.templates/pull/613/commits/ff1f837ca6693a58cb7543b63f1572bdb807ec57

stijnmoreels (changes requested):
The location seems wrong, here.

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -5,7 +5,8 @@
   //#if (Serilog)
   "Serilog": {
     "MinimumLevel": {
-      "Default": "Information"
+      "Default": "Information",
+      "APPINSIGHTS_INSTRUMENTATIONKEY": ""
```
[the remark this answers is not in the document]
vtermaat: oops
https://github.com/arcus-azure/arcus.templates/pull/613/commits/0905a5de817e25997da925b9c02d13cde163bd8a

stijnmoreels (approved):
💃🎉🕺

src/Arcus.Templates.WebApi/appsettings.json
```diff
@@ -3,6 +3,7 @@
   "CertificateSubject": "<your-certificate-subject-name>",
   //#endif
   //#if (Serilog)
```
fgheysels, line 5: What exactly does this `Serilog` conditional define ?
Do we have an option to setup logging without using Arcus.Observability & Serilog ?
If we do, then I think we need to use another conditional to determine if the APPINSIGHTS_INSTRUMENTATIONKEY should be included or not.
If we don't, then I think it makes sense to rename that conditional (in another PR).

fgheysels (approved):

src/Arcus.Templates.WebApi/appsettings.json
stijnmoreels: Yes, that's correct. So we have an logging project option that lets the user decide whether to use the default Console or the Serilog/Arcus AzureApplicationInsights setup. The `APPINSIGHTS_INSTRUMENTATIONKEY` is therefore only used with this second option and can therefore also be placed in the same conditional expression.

fgheysels: Then I think the SErilog conditional should be renamed for clarity :)  (Arcus_AppInsights ? Serilog_AppInsights ? AppInsightsLogging ?)

stijnmoreels: We could maybe discus this in another issue, and not burden Vincent with this 😄 .
We could indeed add some relationship with AppInsights in the directive, but `Serilog` should be part of that, as the project option is adding Serilog to the project. 😉

vtermaat: Hm. Out of those three, `Serilog_AppInsights` makes the most sense to me.
I assume this refers to the same conditional?:

```
#if Serilog
using Serilog;
using Serilog.Configuration;
using Serilog.Events; 
#endif
```

Those usings are very Serilog specific so I'd expect some reference to Serilog to be in there.

fgheysels: > We could maybe discus this in another issue, and not burden Vincent with this 😄 . We could indeed add some relationship with AppInsights in the directive, but `Serilog` should be part of that, as the project option is adding Serilog to the project. 😉

Yes, that's why I also mentionned that we should do it via another PR :)
I'll create an issue for this.

PR #622, opened 2022-07-08
test: add integration test for webapi w/ excluded exception in health report

stijnmoreels:
Adds an integration test that verifies that we indeed removes the exception details when the application's health is unhealthy.


fgheysels (approved):

stijnmoreels:
Due to faulty merge, there was a problem with the JWT Swagger security requirement, but is fixed here https://github.com/arcus-azure/arcus.templates/pull/622/commits/858c1c5b4ecd978ad13d055385d7f10f365aedc7

Issue #620, opened 2022-07-06
Retrieve Application Insights secrets from Arcus' secret store


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
We still get the Application Insights instrumentation key from the application configuration in the project templates because it would otherwise cause a endless loop (secret store requires logging, logging requires secret).

**Describe the solution you'd like**
WIth the `CreateBootstrapLogger` extension on Serilog's  `LoggerConfiguration` and the `.Reload` method it provides, we can reload the logger after the host is run, and so, adding the Application Insights sink with Arcus secret store's secrets.

stijnmoreels:
- [ ] Azure Functions HTTP trigger
- [ ] Azure Functions Service Bus queue trigger
- [ ] Azure Functions Service Bus topic trigger
- [ ] Azure Functions Databricks timer trigger
- [ ] Web API
- [ ] Service Bus queue worker
- [ ] Service Bus topic worker

stijnmoreels:
For Azure Functions, we need some additional features in Arcus libraries to transform them to isolated Functions so we can use the Serilog' reloadable features.

stijnmoreels:
Done as far as we could. Isolated Azure Functions will be able to do this, too.

PR #638, opened 2022-07-20
feat: use newest features in observability/webapi

stijnmoreels:
Use the newest features in observability and web API in the Web API project template.

Relates to #620


src/Arcus.Templates.Tests.Integration/WebApi/WebApiProject.cs
```diff
@@ -191,36 +191,13 @@ public static WebApiProject CreateNew(TestConfig configuration, WebApiProjectOpt
...
-
-            return project;
-        }
-
-        /// <summary>
-        /// Verifies if the specified file is part of the project.
-        /// </summary>
-        /// <param name="filename">The complete path and filename of the file, relative to the root folder of the project (without leading (back)slash).</param>
-        /// <returns>True if the file is present, otherwise false.</returns>
-        public bool ContainsFile(string filename)
```
stijnmoreels, line 204: Is already made available in the parent base class, so no need to implement it here, too.

fgheysels (approved):
Looks good!

PR #665, opened 2022-08-29
feat: add exclude serilog project option to az func http trigger project template

### LINKED ISSUE #598: Add project option to exclude Serilog from Azure Functions HTTP trigger project template
We should not force a certain logging system on consumers and allow others to be used without much effort. In other project templates we added an 'exclude' option to switch to Console logging as default for example, but we didn't do that yet in the Azure Functions HTTP trigger project template.

Adds an `--exclude-serilog` project option to the Azure Functions HTTP trigger project template that will remove any Serilog package and code from the resulting project, similar to the other project templates.

stijnmoreels:
Adds project option to Azure Functions HTTP trigger project template to exclude Serilog as logging system.

Closes #598


fgheysels (approved):
Nice to make this optional

PR #666, opened 2022-08-29
chore: fix open api descriptions

stijnmoreels:
Some correlation ID's were having wrong descriptions.
Fixed that.


fgheysels (approved):
Thanks!

PR #684, opened 2022-09-19
feature/add default http port

### LINKED ISSUE #683: Add `ARCUS_HTTP_PORT` in `appsettings.json` so that default Web API project templates run directly
When the Web API project template is run, the `ARCUS_HTTP_PORT` configuration value is not filled out, which means that it cannot be run by default.

If we would provide such a default value to the `appsettings.json`, then the project would run by-default, without any extra values.

stijnmoreels:
Add default HTTP port to Web API project template in both with and without `appsettings.json`:
- with: provide default HTTP port in `appsettings.json`
- without: provide default HTTP port within code, if none is provided

Closes #683

### DIFF
--- a/src/Arcus.Templates.WebApi/Program.cs
+++ b/src/Arcus.Templates.WebApi/Program.cs
@@ -304,7 +304,11 @@ namespace Arcus.Templates.WebApi
 #endif
         private static void ConfigureHost(WebApplicationBuilder builder, IConfiguration configuration)
         {
-            string httpEndpointUrl = "http://+:" + configuration.GetValue("ARCUS_HTTP_PORT", 5000);
+#if AppSettings
+            string httpEndpointUrl = "http://+:" + configuration.GetValue<int>("ARCUS_HTTP_PORT"); 
+#else
+            string httpEndpointUrl = "http://+:" + configuration.GetValue("ARCUS_HTTP_PORT", 5000); 
+#endif
             builder.WebHost.ConfigureKestrel(options => options.AddServerHeader = false)
                            .UseUrls(httpEndpointUrl);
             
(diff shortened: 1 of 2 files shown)


stijnmoreels:
Can add some tests, but waiting for #682 , otherwise a lot of merge conflicts will be added.

src/Arcus.Templates.Tests.Integration/WebApi/WebApiProject.cs
```diff
@@ -206,10 +206,19 @@ public static WebApiProject CreateNew(TestConfig configuration, WebApiProjectOpt
...
         {
-            Run(_configuration.BuildConfiguration,
-                TargetFramework.Net6_0,
-                CommandArgument.CreateOpen("ARCUS_HTTP_PORT", _baseUrl.Port));
+            var commandArguments = Array.Empty<CommandArgument>();
+            if (ContainsFile("appsettings.json"))
+            {
+                UpdateFileInProject(
+                    "appsettings.json", 
+                    contents => contents.Replace("\"ARCUS_HTTP_PORT\": 5000", $"\"ARCUS_HTTP_PORT\": {_baseUrl.Port}"));
```
fgheysels, line 214: I don't think 5000 is a good idea to use as a custom port.
[Usage of port 5000](https://www.speedguide.net/port.php?port=5000)

It looks to be the default port that is used by Kestrel though.  Idk, maybe we should specify another port that exists in the range 1024 - 49151
https://www.sciencedirect.com/topics/computer-science/registered-port#

Just a suggestion.

fgheysels (approved):

src/Arcus.Templates.Tests.Integration/WebApi/WebApiProject.cs
stijnmoreels: Yeah, that's probably why 5000 was used here, bc kestrel was using it.
But, actually, I don't have any preference here. Fine to use within that range.

stijnmoreels: Changed to 4068.

Issue #414, opened 2021-08-13
Move Azure Functions project templates towards out-of-process functions


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Currently, we still struggle to use the full potential of a safe, secure, functional web API via Azure Functions. Missing middleware, filers, routing and other functionality makes this quite hard.

**Describe the solution you'd like**
Starting from .NET 5, we should be able to have more control over the startup process of the Azure Functions.

**Additional context**
https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide

stijnmoreels:
Since we finally talk about moving towards .NET 6, we could consider moving towards isolated Azure Functions too! 🎉 🥳 
This will also include middleware, which will probably also be helpful when we integrate service-to-service correlation.

PR #685, opened 2022-09-20
feat: add isolated az func servicebus topic project option

stijnmoreels:
Adds a project option to the Azure Functions Service Bus topic project template that lets the user decide whether the project should use the in-process or isolated worker type.

Relates to #414


src/Arcus.Templates.AzureFunctions.ServiceBus.Topic/OrderFunction.cs
```diff
@@ -51,6 +59,45 @@ public OrderFunction(IAzureServiceBusMessageRouter messageRouter)
...
+            }
+
+            var message = ServiceBusModelFactory.ServiceBusReceivedMessage(
+                body: BinaryData.FromBytes(messageBody),
+                messageId: context.BindingContext.BindingData["MessageId"]?.ToString(),
+                correlationId: context.BindingContext.BindingData["CorrelationId"]?.ToString(),
+                properties: applicationProperties);
+
+            return message;
         }
```
stijnmoreels, line 100: Message-specific metadata is not yet supported for isolated functions. Though, they are already in the binding context.
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=isolated-process%2Cextensionv5&pivots=programming-language-csharp#message-metadata

stijnmoreels:
Ah, will probably need to add an additional Docker test here as the project option really affects the project template.

src/Arcus.Templates.AzureFunctions.ServiceBus.Topic/OrderFunction.cs
```diff
@@ -51,6 +59,45 @@ public OrderFunction(IAzureServiceBusMessageRouter messageRouter)
...
+#endif
+#if Isolated
+        /// <summary>
+        /// Process an Azure Service Bus topic <paramref name="messageBody"/> as an <see cref="Order"/>.
+        /// </summary>
+        /// <param name="messageBody">The incoming message on the Azure Service Bus topic, representing an <see cref="Order"/>.</param>
+        /// <param name="context">The execution context for this Azure Functions instance.</param>
+        [Function("order-processing")]
+        public async Task Run(
+            [ServiceBusTrigger("order-topic", "order-subscription", Connection = "ServiceBusConnectionString")] byte[] messageBody,
```
fgheysels, line 72: Just wondering why we receive a byte array here, instead of a ServiceBusReceivedMessage, like we have in the in-process version ?

fgheysels (approved):
Looks good to me, just a simple question for something I'm wondering about

src/Arcus.Templates.AzureFunctions.ServiceBus.Topic/OrderFunction.cs
stijnmoreels: Because that is not supported. The isolated variant only allows access to the message's contents and via via a limited access to some (but not all) metadata.
https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=isolated-process%2Cextensionv5&pivots=programming-language-csharp#message-metadata

PR #712, opened 2022-11-10
chore: remove installation details from readme

### LINKED ISSUE #640: Update README with available project templates
Currently, we only list the Web API and the Azure Service Bus worker project templates.

List all the available project templates, or point them directly to our feature documentation, as it would become a rather long list if we would list (and maintain) them all in the README file.

stijnmoreels:
It is hard to maintain and keep up to date and it does not provide enough added-value as it is a duplication of the documentation, so removed the installation details from the `README.md` file.

Closes #640

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,49 +6,7 @@
 
 ![Arcus](https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png)
 
-# Installation
-
-## Web API Template
-We provide a template to build Web APIs. ([docs](https://templates.arcus-azure.net/features/web-api-template))
-
-First, install the template:
-```shell
-> dotnet new --install Arcus.Templates.WebApi
-```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,49 +6,7 @@
 
 ![Arcus](https://raw.githubusercontent.com/arcus-azure/arcus/master/media/arcus.png)
 
-# Installation
-
-## Web API Template
-We provide a template to build Web APIs. ([docs](https://templates.arcus-azure.net/features/web-api-template))
```
fgheysels, line 12: The advantage of this was, that you could see this info immediately when opening the project on github, without having to look into the separate docs.
But I can agree that it is duplicate info and therefore also hard to maintain

fgheysels (approved):

README.md
stijnmoreels: Yes, that's the reason for this proposal. I think that even all `README` files were out-of-date.

Issue #764, opened 2023-01-12
Streamline `launchSettings.json` in project templates


stijnmoreels:
**Is your feature request related to a problem? Please describe.**
Currently, we have some project templates that have a `/Properties/launchSettings.json` file and others don't. This is not consistent.

**Describe the solution you'd like**
Make sure that every project template has a `/Properties/launchSettings.json` file that both works with Docker and the project itself, so developers can both easily start the project via Docker or by itself.

fgheysels:
This is only relevant for Web API project templates, isn't it ?

stijnmoreels:
> This is only relevant for Web API project templates, isn't it ?

No, there are project templates (like the Service Bus topic worker) that do not have such settings file, while others (like the Service Bus queue worker) has.

PR #765, opened 2023-01-12
fix: do not exclude launchsettings.json on webApi template

Roeland54:
launchsettings.json was not added to the webapi project by this.

Relates to #764


stijnmoreels:
Thx @Roeland54 for your PR! 🎉 Was indeed we were planning to streamline in #764 for all our project templates.

stijnmoreels (approved):

fgheysels (approved):
Thanks for contributing!

stijnmoreels:
On second thoughts, I see that this does not take effect in the availability of the `launchSettings.json`. Possibly was tested on the v1.0, while our recent `main` branch already had this fix; regardless of this PR. In any case, it's more clear this way.

PR #801, opened 2023-04-11
chore: add az func local settings

### LINKED ISSUE #782: Improve Function Templates by configuring Azurite
When you create a new Azure Function, you typically also want to run it locally during development.  
Since the Azure Storage Emulator has been discontinued and is replaced by 'azurite', we should configure our projects to support azurite.

For more information, see [this documentation](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#running-azurite-from-an-azure-functions-project).

I believe we can simply do this by making sure that the `local.settings.json` file  contains this entry:

```json
"AzureWebJobsStorage": "UseDevelopmentStorage=true"
```

stijnmoreels:
Add `local.settings.json` file with Azurite support for every Azure Functions project template.

Closes #782

### DIFF
--- a/src/Arcus.Templates.AzureFunctions.Databricks.JobMetrics/local.settings.json
+++ b/src/Arcus.Templates.AzureFunctions.Databricks.JobMetrics/local.settings.json
@@ -1,13 +1,10 @@
 {
-    "IsEncrypted": false,
+  "IsEncrypted": false,
   "Values": {
     //#if(Isolated)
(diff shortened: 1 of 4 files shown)


fgheysels (commented):
Small but really helpfull.

Sounds like 'Vermaelens Projects'.

fgheysels (approved):

PR #867, opened 2024-05-02
feat: remove in-process functions worker project option

stijnmoreels:
Remove the project option to create an in-process Azure Functions project, since the .NET 8 framework does not support in-process. Also, remove the Databricks project since this was only supported for in-process models.

Relates to https://github.com/arcus-azure/arcus/issues/311


stijnmoreels:
Bear with me, it seems like a lot of changes, but they are almost all deletions. 😅

stijnmoreels:
Fails on Azure Functions EventHubs integration test, but also fails on `main`. Only thing changes there is that it uses a different `func` local tool since that was updated.

fgheysels (approved):

Issue #854, opened 2023-12-04
base-images for runtime in Dockerfiles need to be reviewed


fgheysels:
When I create a new project based on the Arcus Messaging template, I can see that the generated dockerfile looks like this:

```
FROM mcr.microsoft.com/dotnet/sdk:6.0.301-alpine3.14 AS base
WORKDIR /app
EXPOSE 80

FROM mcr.microsoft.com/dotnet/sdk:6.0.301-alpine3.14 AS build
WORKDIR /src
COPY ["Arcus.Templates.ServiceBus.Topic.csproj", ""]

COPY . .
WORKDIR "/src/."
RUN dotnet build "Arcus.Templates.ServiceBus.Topic.csproj" -c Release -o /app

FROM build AS publish
RUN dotnet publish "Arcus.Templates.ServiceBus.Topic.csproj" -c Release -o /app

FROM base AS final
WORKDIR /app
COPY --from=publish /app .
ENTRYPOINT ["dotnet", "Arcus.Templates.ServiceBus.Topic.dll"]
```
So, this means that the base-image that is used for running the project, is an SDK image.  This is not necessary.  We must use a 'runtime' image, like f.i. ` 6.0.25-alpine3.18-amd64`

We must review all Dockerfiles in all templates to make sure that no 'sdk' versions are used for running our projects.

fgheysels:
Shall we do this together with the .NET 8 release ?

stijnmoreels:
> Shall we do this together with the .NET 8 release ?

We could yes.

PR #873, opened 2024-05-07
chore: upgrade templates to net8.0

### LINKED ISSUE #869: Investigate if we can remove the pinned version of Azure Functions CLI tools
Make sure that we remove back the pinned Azure Functions CLI tool in the PR #867 while we upgrade towards .NET 8.

stijnmoreels:
Upgrade project templates to .NET 8.

Relates to https://github.com/arcus-azure/arcus/issues/311
Closes #869


stijnmoreels:
Sorry for the big PR, limited only to package updates and breaking changes fixes.

src/Arcus.Templates.AzureFunctions.EventHubs/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/azure-functions/dotnet:4 AS base
 WORKDIR /app
 EXPOSE 80
 
-FROM mcr.microsoft.com/dotnet/sdk:7.0.203-bullseye-slim AS build
+FROM mcr.microsoft.com/dotnet/sdk:8.0-bookworm-slim AS build
```
stijnmoreels, line 5: Not sure if these are correct, seems like the most simple ones. I like simple 😅.

build/templates/docker-integration-tests.yml
```diff
@@ -14,10 +14,6 @@ stages:
             inputs:
               artifact: 'Templates'
               path: '$(Build.SourcesDirectory)/templates'
-          - task: DotNetCoreInstaller@0
```
stijnmoreels, line 17: Is actually not needed to run the tests.

build/ci-build.yml
```diff
@@ -89,8 +89,6 @@ stages:
         steps:
           - task: FuncToolsInstaller@0
             displayName: 'Install Azure Functions Core tools'
-            inputs:
-              version: '4.0.5390'
```
stijnmoreels, line 93: Removed this pinned version again, and works now with the latest CLI tool and .NET 8.

src/Arcus.Templates.Tests.Integration/AzureFunctions/ServiceBus/AzureFunctionsServiceBusProject.cs
```diff
@@ -134,17 +134,14 @@ public class AzureFunctionsServiceBusProject : AzureFunctionsProject, IAsyncDisp
...
         {
-            AddPackage("Arcus.EventGrid.Core", "3.3.0");
-
             AddTypeAsFile<Order>();
             AddTypeAsFile<Customer>();
-            AddTypeAsFile<OrderCreatedEvent>();
             AddTypeAsFile<OrderCreatedEventData>();
 
-            AddTypeAsFile<TestOrdersAzureServiceBusMessageHandler>();
+            AddTypeAsFile<WriteToFileMessageHandler>();
```
stijnmoreels, line 141: Used a `WriteToFileMessageHandler` here. Tt actually makes the integration tests a bit more stable, as we do not need to interact with EventGrid anymore. We inject a message handler that writes the message on disk instead and look for such a file during the assertion. It also removes the need to deploy such an EventGrid resource for the Templates repo + it was actually unnecessary additional overhead to bring EventGrid in the loop here, as we only need to test if the generated project is capable of routing messages.

src/Arcus.Templates.AzureFunctions.EventHubs/Dockerfile
fgheysels: I find those names always confusing, but it might be me :)

fgheysels: Bookworm seems to be debian (12).
Bullseye is also debian (11), so that's ok.

src/Arcus.Templates.Tests.Integration/AzureFunctions/ServiceBus/AzureFunctionsServiceBusProject.cs
fgheysels: Indeed, if we only need to test the routing, no need to add additional complexity and dependencies.  I like it simple :)

fgheysels (approved):

src/Arcus.Templates.AzureFunctions.Http/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/azure-functions/dotnet:4 AS base
 WORKDIR /app
```
fgheysels, line 2: Is dotnet:4 the correct image that we should use as runtime base image ?

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM mcr.microsoft.com/azure-functions/dotnet:4 AS base
 WORKDIR /app
```
fgheysels, line 2: Is this still the correct base image for runtime ?

src/Arcus.Templates.EventHubs/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0.301-alpine3.14 AS base
+FROM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS base
```
fgheysels, line 1: The runtime base image that must be used, should not be an sdk version.
We should use one of [these](https://hub.docker.com/_/microsoft-dotnet-runtime/).

I would suggest mcr.microsoft.com/dotnet/runtime:8.0-alpine

See also https://github.com/arcus-azure/arcus.templates/issues/854

src/Arcus.Templates.ServiceBus.Queue/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0.301-alpine3.14 AS base
+FROM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS base
```
fgheysels, line 1: The runtime base image that must be used, should not be an sdk version.
We should use one of [these](https://hub.docker.com/_/microsoft-dotnet-runtime/).


I would suggest mcr.microsoft.com/dotnet/runtime:8.0-alpine

Using an SDK image as the base-image for the runtime version of the container is too bloated.  It contains all kind of stuff that you don't need during runtime.  

See also https://github.com/arcus-azure/arcus.templates/issues/854

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0.301-alpine3.14 AS base
+FROM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS base
```
fgheysels, line 1: The runtime base image that must be used, should not be an sdk version.
We should use one of [these](https://hub.docker.com/_/microsoft-dotnet-runtime/).


I would suggest mcr.microsoft.com/dotnet/runtime:8.0-alpine

Using an SDK image as the base-image for the runtime version of the container is too bloated.  It contains all kind of stuff that you don't need during runtime.  

See also https://github.com/arcus-azure/arcus.templates/issues/854

src/Arcus.Templates.AzureFunctions.Http/Dockerfile
stijnmoreels: I believe that the `dotnet:4` is about the v4 version of Azure Functions, so yes.

stijnmoreels: Will be looked over in a follow-up PR.

src/Arcus.Templates.EventHubs/Dockerfile
stijnmoreels: Will be looked over in a follow-up PR.

src/Arcus.Templates.ServiceBus.Topic/Dockerfile
stijnmoreels: Will be looked over in a follow-up PR.

src/Arcus.Templates.ServiceBus.Queue/Dockerfile
stijnmoreels: Will be looked over in a follow-up PR.

src/Arcus.Templates.AzureFunctions.ServiceBus.Queue/Dockerfile
stijnmoreels: Will be looked over in a follow-up PR.

PR #879, opened 2024-05-23
chore: upgrade package/docker dependencies

### LINKED ISSUE #854: base-images for runtime in Dockerfiles need to be reviewed

### LINKED ISSUE #878: Investigate if we can use 'chiseled' .NET base images in our Dockerfiles
We should investigate if we could use chiseled images as base images in the Dockerfiles that we integrate in our Arcus templates.

Chiseled images are lightweight and reduce the attack surface.
https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8/containers#chiseled-ubuntu-images

stijnmoreels:
* Minimize the used `Dockerfile`s with the new .NET 8 upgrade, both with using `runtime` instead of `sdk` and using explicitly `dotnet-isolated` for Azure Functions & chiseled for building,
* Upgrade all nuget packages to latest version,
* Remove all leftover Databricks project files.

Closes #854 
Closes #878


src/Arcus.Templates.ServiceBus.Queue/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM mcr.microsoft.com/dotnet/sdk:8.0-alpine AS base
+FROM mcr.microsoft.com/dotnet/runtime:8.0.4-jammy-chiseled AS base
```
fgheysels, line 1: Just wondering why the runtime image is a 'jammy' image (Ubuntu 22.04) while the base image that we use for the build phase, is a bookworm (Debian based) image ?
https://github.com/dotnet/dotnet-docker/blob/main/README.runtime.md#linux-amd64-tags

stijnmoreels: Ok, changed all to `jammy` in places where we use `chiseled`.

fgheysels (approved):

PR #880, opened 2024-05-28
chore: use nullable option in project templates

### LINKED ISSUE #865: Improve the definition of the ApiHealthReportEntry class

When creating a new project based on the Arcus.WebApi template, we might get compiler warnings because of this line:

```csharp
 return new ApiHealthReportEntry
            {
                Data = entry.Data.ToDictionary(item => item.Key, item => item.Value),
                Description = entry.Description,  // here we might have compiler warnings
                Duration = entry.Duration,
                Status = entry.Status,
                Tags = entry.Tags
            };
```

The compiler is issued because it is possible that we assign null on a non-nullable property.
Indeed, we see that the `Description` is declared as a `string` whereas the `Description` property of the `HealthReportEntry` class is declared as a `string?`: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.healthreportentry.description?view=dotnet-plat-ext-8.0

I think we should declare the `Description` property of the `ApiHealthReportyEntry` as a nullable string as well.

stijnmoreels:
Use the `<nullable>true</nullable>` option in the projects, and fixes any possible `null` assignment.

Closes #865


fgheysels (approved):
Great ! 
I like this stricter approach for nullable types

PR #895, opened 2024-07-09
chore: remove eventgrid reference from eventhubs testing

stijnmoreels:
Remove EventGrid reference from EventHubs integration testing and user local disk as event source.


stijnmoreels:
@fgheysels , best to check this PR before other ones, as it will greatly improve the stability of parallel integration test runs.

fgheysels (approved):

